PR #3, opened 2015-04-26
Create Mac version

cochrane:
This creates a Mac project and changes enough things to get it working on Mac OS X again, without changing anything for other platforms. Important changes:

-   The SDL framework is included.
-   The FindConfigFile files are included again (they were referenced
in engine.cpp but not present anymore).
-   In gl_util, the code no longer tries any funny business to get
OpenGL includes working on Macs, and just imports SDL/opengl.h directly
on all platforms. This should work on other platforms, too, but if
anything breaks other platforms, it’s probably this.
-   SDL_image and the included copies of libpng and libjpg are not used
(because getting them to work was way too much effort). Instead
standard Mac OS X methods are used to read either kind of file. This
means the Mac version can also read some file formats such as PSD that
the Windows version can’t, though this will probably not make a
difference in practice.


richardba:
Done! Thanks!

PR #19, opened 2015-05-09
Bugfix: Small fix for forward slide state.

Gh0stblade:


Lwmte:
Actually, the way forward slide is processed in TR3+ is purposeful, as there is unused slide-to-run animation. I never seen problems with that, and it smooths out gameplay a bit, so why not leave it?

Gh0stblade:
Oh I was no aware of this unused animation. I just thought it looked a little odd being able to run directly after sliding down the slope! Closing this! Thanks.

PR #20, opened 2015-05-10
Lua error handling

cochrane:
Lua errors were previously simply ignored. That resulted in bugs not being found, and a few crashes due to lua code being called in incorrect ways. To solve this, this merge requests adds a new function lua_CallAndLog, which is meant to be used instead of lua_pcall. It calls lua_pcall internally. If any error occurs, it prints it to the console and deals with the stack. It also logs where in C code the function was called. This should make debugging of lua scripts quite a bit easier. Having it all in one place also means that further changes to debugging code are much easier to make.


cochrane:
Additionally, now includes all the matrix cleanup that I talked about on the forum.

Lwmte:
Hmm, it seems there is some garbage in script.cpp, I'll try to fix that manually.

PR #54, opened 2015-06-05
[Add] Configurable Brightness

Gh0stblade:


Lwmte:
Hmmm... I tried it, but seems that brightness setting has no effect on my machine?

Gh0stblade:
Strange, it does work for me. Did you also update the config.lua in your binary folder? It's defaulted to 1.0 unless modified. If you did and it still did not work. Please close this pull request.

Lwmte:
Yes, I double-checked it, and it really doesn't work! On both GeForce 635M and Intel-whatever integrated video. Strangely, **SetWindowGammaRamp** doesn't work as well! I believe it may be the problem of SDL2. I'll post a bug report to them, when I have time.

vvs-:
Intel Linux drivers have big problems with setting gamma. And to be fair, I've seen such problems on other cards as well. Just my two cents.

PR #104, opened 2015-06-30
CXX-ification of OpenTomb

stohrendorf:
This is an attempt to clean up the code base and make use of C++ features. **Bells and Whistles here**, because this is a huge refactoring, and despite all care taken, I could have introduced some bugs, although I think this improves maintainability in the long term.


cochrane:
Wow, that's an impressive amount of work! This will probably need some discussion, but I for one am in favor. I've written a longer text about it here: http://www.tombraiderforums.com/showthread.php?p=7390202#post7390202

Not sure where the bulk of the conversation about this will take place; reply wherever you prefer.

vvs-:
Removing in-tree libraries has the potential to introduce subtle incompatibilities, especially on different platforms. I believe that Linux will have an advantage here, but most developers and users are on "that other OS".

cochrane:
I'm not that concerned about incompatibilities. All these libraries have been around forever and are considerably better debugged than OpenTomb. But we do need to figure out what we do instead. Do we put the libraries for Windows and Mac into the repository (my preference, and IIRC all of these are at least LGPL so doing that is no problem), add them as external Git repositories (never been very fond of that), or something?

I am somewhat concerned about OpenAL. My default approach would be to revert to the version that ships with Mac OS X, but that lacks some of the features that OpenTomb currently uses. I guess I could use a custom version, though, as far as I know most games do that.

Gh0stblade:
I have no problem with this. Anything contributing to cleanups is a huge+ imo. Of course, there is a risk of new bugs being created.

stohrendorf:
@cochrane I think I will stay here first, until this discussion gets too complex.

@vvs- I think this is a wrong assumption; having copies of libraries in-tree means taking responsibility of compiling them for different platforms, effectively removing the benefits of having the respective packagers dealing with platform subtlenesses. Even on Linux, nearly every distro has a few patches laying around for some libraries, which you really don't want to maintain ;)

@cochrane If you ever want OpenTomb to be released, you have to make sure you are using a library version that's available on all target platforms, meaning that you should stick to the version you're currently developing with and wait until it gets released on all target platforms.

@Gh0stblade Yep, I'm pretty sure that I have introduced a small bug here or there...

vvs-:
As I already said it's not me who might object.

vvs-:
> Do we put the libraries for Windows and Mac into the repository

Oh no! Just not that. Keeping binaries in a repository is a bad idea. I don't know how many times I swear when forced to download the countless huge blobs on each update.

cochrane:
Okay. I think I can use Cocoapods to make the process painless and simple for Mac developers. No idea about other platforms, but I'm sure they can figure something out (especially Linux)

Lwmte:
First of all, thank you SO MUCH for such impressive work you have done!
I need some time to learn the meaning of some changes, cause on the first look, I have completely no idea what's happening here or there! =)

CXX-fication of code was always a controversial subject here, as TeslaRus intended to make code as much plain-C as possible, because of performance issues - however, I for myself don't know if std/boost code also performs slowly than plain-C. Overall, CXX code looks more intuitive and is shorter than equal C code, but sometimes, exactly opposite thing is happening. For example, when this

` Con_AddLine("blah"); `

becomes this:

` ConsoleInfo::instance().addLine("blah"); `

But again, I just took a very quick look at parts of the code, I need to learn it more to get an idea why you did this or that change.

Considering getting rid of library copies, i *completely* agree with this decision, *but* when it comes to Lua, Ogg+Vorbis and Freetype. As for Bullet - I know that TeslaRus made some changes to Bullet code, particularly in debug drawer, so there was a point keeping it with the source code. As for OpenAL - I'm not sure (yet) how you've ruled out this, I hope we're not downgrading to vanilla OpenAL, nor we're going to use plain OpenAL Soft 1.15.1? There was an issue with low-pass filtering in click-removal code, that was the whole point of using bundled OpenAL copy; also, neither OpenAL (vanilla) nor OpenAL soft uses SDL backend, which was custom-written... However, I don't see a difference here, if we won't suffer performance or compatibility issues... actually I've accidentally used external OpenAL lib without even noticing it for some time.

stohrendorf:
I don't want to sound like I'm owning this project now, but from my personal experience it is *always* better to first write clean, easily-understandable code (i.e. the good old KISS principle), and then *first* try to do macro-optimizations (i.e. algorithmic changes) before doing micro-optimizations.

About the coding style: This isn't a core library, so I don't see the need to stick to that style guide (which, btw, is a legacy thing from C and when computer only had uppercase letters); a prefix of "m_" in classes tells the reader that this isn't a global variable; the casing also helps to distinguish between "TypeNames", "m_emberVariables", "MACROS", etc.; camelcase also reduces more underscores than the "m_" adds; and last but not least the style I'm using is also used by many huge and important frameworks/libraries.

@Lwmte So the libs got patched, making them forks of the original ones? Oh dear, let's release the "TROpenAL", "TRLua", "TRBullet", "TROGG" and "TRVorbis" libs to the world :grimacing:

vvs-:
Yes, C++ is more verbose, but C is unsafe. As it was C from the beginning then there is no other choice left.

cochrane:
I fully agree with @stohrendorf on performance. Performance problems don't happen where you expect them anyway, they occur somewhere you never would have thought to look without a profiler. Using proper C++ is pretty much guaranteed not to hurt performance if you don't use exceptions and virtual methods (which this code doesn't as far as I can tell), and even if you do it's incredibly unlikely to be a problem. Writing proper C++ code, instead of a weird version of C89 with .cpp file suffixes, is definitely the right move. It can decrease readability if you go overboard (such as almost any non-trivial use of templates), but if used properly, it makes the code overall better to read. This pull request seems to do that.

@stohrendorf: I don't want to spend too much time arguing about the `m_`; if I'm the only one who cares, then I'm perfectly fine with keeping it. But on a philosophical level, I believe that if you have so many global variables that having that `m_` actually helps, then something has gone very wrong. :D

I don't know anything about the patching story (but that would explain the weirdness I've seen when trying to use normal versions of these libraries). If these patches are actually required, and I hope they're not, then I'd strongly suggest to put these into new repositories (probably under the OpenTomb organization as well) so we get a clear separation of responsibilities. And also so others can use it if they feel the need.

stohrendorf:
Just as a side note: This refactoring has only started yet, it's at most 15% done. I'm doing this pull request because it got more and more troublesome to merge your master into the code. I'm planning to do even more refactoring, but not until this pull request has been merged and the first tests have been done to identify introduced problems.

(Let's put the coding style discussion aside now...)

vvs-:
First impressions. The `engine.cpp` is really slow to compile and produce a warning. I really hate the C++ templates' ability to be a resource hog. Another warning in `entity.cpp` and it's about missing bullet API. And it doesn't work because of segmentation faults. Not sure what's a culprit here. It could be a library problem.

Lwmte:
As I remember, changes to Bullet were not critical, mainly related to performance issues (profiling was enabled in Bullet by default, and patch changed that).
For good news, my lowpass filter patch isn't actual anymore in latest version of OpenAL Soft (1.16.0), I just checked it, there's no sign of lowpass, it was replaced with highshelf, as it was in my patch.

To make things clear: **only patched libs** were Bullet (need to ask TeslaRus to which extent he patched it) and OpenAL (SDL backend support, not approved by Sam Latinga). SDL_Image had some modifications, but we had a talk with TeslaRus recently about removing that as well.

To ease everything for you, I think I'll stop any modifications to core source code until merge, I'll maybe just update some scripts... Also we need some kind of approval from TeslaRus before merge, so let's wait what he'll say on this subject.

TeslaRus:
hmmm... if engine written on C++ than it is correct to use C++ architecture / language; if C++ features are used for task solving - it is good, C++ for C++ is evil. In my opinion best example of good C++ using in game engine is ID software code: doom3, quake4, a.t.c. 
This time only one library are changed for OpenTomb engine: OpenAL soft (used SDL backend);  This time bullet has no my debug drawer optimizations, only profiling are disabled. Bullet may be needed inside project  in case of changing btScalar (float / double), but not sure it can happens; All other libraries may be really deleted from project;
About code change: Really big and great work! about pull request: ok, but please, without boost (too many libraries are used and boost is too heavy for that engine). 
+ I did not test that code: what about fps change / something was broken?

TeslaRus:
@vvs-  "I really hate the C++ templates' ability to be a resource hog." me too =)

stohrendorf:
@TeslaRus boost is used header-only for generating the lua/c++ wrappers in luahelper.h; and BTW, templates may slow down the compilation time, but the runtime performance hit is neglectible, it can even improve performance due to type-checked optimizations. I will try to reproduce the crash now and make the fixes if I find them...

godmodder:
As far as I can see there is not much templated code in engine.h/cpp. I suspect including headers of other libraries like bullet in .h files is hurting compile performance exponentially here. It's unfortunate that there's no standard way to use precompiled headers, so there's not much more that we can do than keep the code as untangled as possible. Maybe the PIMPL idiom could help here. That and doing as much in .cpp's as possible and forward declare wherever we can.

I agree with stohrendorf about the templates: there is no runtime overhead, only compile overhead. Use templates wherever it makes sense and where it shortens code significantly.

TeslaRus:
about templates: If "there is no runtime overhead, only compile overhead" than all are ok.
Probably I will need to deal with advanced C ++ in work, so it's a sign from above; 
do merge - ok, heh, but boost... ahmm, ok

stohrendorf:
Too bad it's already merged... I just wanted to say that I found some very bad design issues I'm working on right now...

vobject:
Having libraries inside the repo made it much easier to setup a development environment. Since we have automated build this has become much less of an issue as users who just wanted to have a peek at the latest state don't have to compile it themselves. Still, we need a documentation about how to setup the development environment. On Linux, this is very easy, but on Windows you can have a lot of issues with installing (and deploying!) all the libraries for MinGW.

Going for C++ standard library types is a great move as there are a lot of memory related errors that come from writing every little loop by hand: malloc->delete[] mistakes, aliasing/ memory overlapping issues etc.

As for boost: please let's try to avoid it for now. At least the non-header-only libs.

vvs-:
Templates are apparently coming from boost and not from OpenTomb itself.

Another note: memory usage has doubled. But it could also be attributed to a crash.

Lwmte:
Also I've noticed some wrong variable/function renamings, like **light_position** became **Lightosition** (missing P) or **sprite_texture_count** is now **Spriteexture_count** (missing T). Please be careful with that! :)

TeslaRus:
@stohrendorf : Too bad it's already merged... I just wanted to say that I found some very bad design issues I'm working on right now...
May be, but do it later is bad too. 
About libs: some libs (if it will be really necessary) may be returned to the repo;
About code style and architecture: we need to describe and confirm some standard for project to avoid multistyle variable / functions / classes names and code style format.

cochrane:
Quick note: I saw the changes also deleted ´src/mac/FindConfigFile.h / .m´. While I'm not sure that either file is the best way to solve the problems I've had, I do need them for the Mac version and plan to add them back in. Does anyone object to that, or have a better idea about where I should put them instead? Putting that code into cross-platform C++ files with conditional compilation is sadly not an option (the code is written in Objective-C instead, because the APIs it accesses require that).

stohrendorf:
Found a serious problem (which was already a bug before): in entity.cpp:157 are setUserIndex *and* setUserPointer used, which set values in a *union*, and then the index (0x0e in this case) gets dereferenced as a pointer in engine.h:162.

vvs-:
Crash happened in gl_font.c(630) working in the console.

Also, no scene is rendered. I still suspect that it's a library issue. I'll try to compile libraries from the old in-tree code and see if anything changed.

vvs-:
With the old in-tree libraries it seems to enter an infinite loop.

Also, somebody please rename `string.h` to something else. It conflicts with C headers.

stohrendorf:
Feel free to create many, many issues in the issue tracker and assign them to me, I will look into them when I'm back home.

Lwmte:
By the way, can we *override* lua print function in some way instead patching print function itself?

stohrendorf:
http://stackoverflow.com/a/4508254 :smiley:

vvs-:
Actually it always was the case. I think that Lwmte actually meant `printf` function, which was patched in Lua.

PR #112, opened 2015-07-02
External lib refactoring

stohrendorf:
The local sources of bullet always leads to a crash at startup as they have different headers than the system libs my build actually links to; the internal 3rd-party sources are only used now when the system libs cannot be found. Additionally, I have removed the need for SDL_Audio and replaced it with libsndfile, which makes the custom openal source obsolete. gl_utils should now use proper API linking, and glext functions are dynamically loaded.


TeslaRus:
GLAPI instead extern... may be it will solve some annoying problems;

Issue #115, opened 2015-07-03
Handling of external libraries


cochrane:
We need to converge with this discussion somewhere because right now, it is all over the place. So here are the possibilities for handling the dependencies our code has that I've been able to think of:

-   Store the code directly in the repository. Pro: Very easy; no setup required; inherently works on all platforms. Can make own modifications. Cons: Can make own modifications that aren't visible to the outside world; update difficult.
-   Store the binaries directly. Pro: Update becomes easy. Con: Need at least three versions of everything (Mac, Linux, Windows), possibly more depending on 32/64 bit. Any new platform (does FreeBSD have OpenGL?) increases the pain.
-   Hope the user has installed the libraries. Pro: Nothing to do on our part except edit the Readme.md. Con: No idea who uses what version of what library.
-   Use external Git references to the libraries (to known tags). Pro: Using official version; no modifications make it back. Same ease of use as storing in repository. Con: Git external references are really hard. Especially if some library we want doesn't have a git repository; there's some Git/SVN bridge but I've never tried it.

A few more notes:
-   SDL is funny because the Mac version is a different format (Framework) than any other library. That's my duty to figure out, of course, for everyone else that's not an issue.
-   How big a problem is incompatibility really? Probably not that big; incompatible OpenGL drivers hurt way more. But of course it takes only one issue to make the whole thing less fun.
-   We definitely need some way to deal with our own version of OpenAL. It does things the normal version doesn't, and I don't want to throw that away without very good reason. We could put it in a separate repository here and then use it like any other third-party library, of course.
-   My personal vote would be for Git external references, I just have very little experience with actually using them.

Lwmte:
Cochrane, is it possible to install/compile/link OpenAL Soft (http://kcat.strangesoft.net/openal.html) for Mac along with stock OpenAL? Bundled version isn't very different from default OpenAL Soft, except it used custom-written SDL backend (which I will really miss - but I wrote a letter to Chris Robinson asking if he can implement SDL backend officially, or at least XAudio2 output for Windows).

[...]

PR #122, opened 2015-07-03
use libsndfile for ogg loading

stohrendorf:
Related to #115

### DIFF
--- a/src/audio.cpp
+++ b/src/audio.cpp
@@ -1,13 +1,9 @@
-
 #include <SDL2/SDL.h>
 
 #include <AL/al.h>
 #include <AL/alext.h>
 #include <AL/efx.h>
 #include <AL/efx-presets.h>
-
-#include <ogg/ogg.h>
-#include <vorbis/vorbisfile.h>
 
 #include "audio.h"
 #include "console.h"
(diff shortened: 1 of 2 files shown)


Lwmte:
Judging by the code, does it mean that streaming is now possible for *any* file format supported by libsndfile?! If that's the case, we can now safely implement CDAUDIO.WAD support!

stohrendorf:
Yes, this is potentially possible, though the main reason for this was to get rid of the SDL-patched openal code.

vvs-:
OpenTomb now hangs after loading a level with repeated message `Attempt to read a non-integer number of channels`.

Lwmte:
Ok, I understand, I won't touch any pull requests anymore until code gets stable! =)

stohrendorf:
Should be fixed now. At least the TR3 soundtrack plays in the background for me.

vvs-:
> Ok, I understand, I won't touch any pull requests anymore until code gets stable! =)

It'd be perfectly viable if there were several branches. A recommended practice is to have an unstable `devel` and stable `master`.

stohrendorf:
> It'd be perfectly viable if there were several branches. A recommended practice is to have an unstable devel and stable master.

I fully agree. I'm currently only fixing small things in `master` and doing larger fixes and every functional change in a separate branch.

PR #155, opened 2015-07-12
Optional FIND_PACKAGE for SDL, for systems without pkgconfig

ablepharus:
On windows without Mingw, there is no pkgconfig. With these changes, CMake can find SDL2 without it.


CMakeLists.txt
```diff
@@ -7,9 +7,14 @@ CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
...
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -msse3")
+IF( ${CMAKE_COMPILER_IS_GNUCC} )
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -msse3")
+ENDIF()
+IF( ${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -msse3")
+ENDIF()
 
-add_subdirectory(3rdparty)
+# add_subdirectory(3rdparty) there is no CMakeLists.txt in 3rdparty
```
stohrendorf: There is :)

ablepharus: But, but....
https://github.com/opentomb/OpenTomb/tree/master/3rdparty

stohrendorf: OK, let's blame @Lwmte for not taking a proper look at the old CMakeLists.txt in d36177d28f693fb53440af5e6e3349f5cac8984a :)

CMakeLists.txt
```diff
@@ -7,9 +7,14 @@ CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
 
 PROJECT(OpenTomb)
 
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -msse3")
+IF( ${CMAKE_COMPILER_IS_GNUCC} )
```
stohrendorf: Remove `${` and `}` to avoid doubled evaluation.

Lwmte:
What actually happened is I forgot to push CMakeLists.txt changes, now I lost my modified CMakeLists.txt... Anyway, it wasn't a big deal, I'll just edit "main" CMakeLists.txt.

By the way, what's up with SDL2/SDL2_Image finders? You mean CMake doesn't include them by default? How it's possible, SDL2 was out for almost 2 years now, and nobody integrated it to CMake yet?

CMakeLists.txt
```diff
@@ -7,9 +7,14 @@ CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
...
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -msse3")
+ENDIF()
+IF( ${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
```
stohrendorf: you can combine both IFs with an `OR`. And enclose the `${...}` with double-quotes.

CMakeLists.txt
```diff
@@ -88,9 +93,19 @@ IF(MINGW)
     ENDIF()
 ELSE()
     INCLUDE(FindPkgConfig)
-    PKG_SEARCH_MODULE(SDL2 REQUIRED sdl2)
-    PKG_SEARCH_MODULE(SDL2_IMAGE REQUIRED SDL2_image)
-
+    IF(${PKG_CONFIG_FOUND})
```
stohrendorf: double-evaluation as said above

CMakeLists.txt
ablepharus: Very nice, the double evaluation was the reason, why I was not able to combine both cases.

CMakeLists.txt
ablepharus: will fix

CMakeLists.txt
ablepharus: 1. combining both ifs - done
2. enclosing ${...} with double quotes - If I understand you correctly:

IF( CMAKE_COMPILER_IS_GNUCC OR  "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" )

Is what you want, but i have to take out the double quotes around CMAKE_CXX_COMPILER_ID because:

" Policy CMP0054 is not set: Only interpret if() arguments as variables or keywords when unquoted."

vvs-:
> You mean CMake doesn't include them by default?

He means that it doesn't work without `pkgconfig` and it's missing on Windows by default.

stohrendorf:
BTW, the reason for being missing in CMake standard modules is because they don't want to maintain external packages (and FindSDL2.cmake is part of SDL2 if you think about it). Hey, why do I want to remove external libs again? :wink:

Lwmte:
SDL2/SDL2_Image were never considered to be submodules in the first place... :yum: 
But this is rather disturbing, I thought that CMake is simple enough just to feed it with CMakeLists.txt and it does everything for you, and now it seems it doesn't.

CMakeLists.txt
vvs-: It's a bad idea trying to introduce new features and fix bugs in the same commit. We already have enough headache with this.

vvs-:
There is no such thing as a free lunch...

CMakeLists.txt
ablepharus: Sorry

PR #167, opened 2015-07-16
Trosettastone

stohrendorf:
asciidoc conversion of trosettastone for further improvability.

### DIFF
--- a/doc/images/icons/README
+++ b/doc/images/icons/README
@@ -1,0 +1,5 @@
+Replaced the plain DocBook XSL admonition icons with Jimmac's DocBook
+icons (http://jimmac.musichall.cz/ikony.php3). I dropped transparency
+from the Jimmac icons to get round MS IE and FOP PNG incompatibilies.
+
+Stuart Rackham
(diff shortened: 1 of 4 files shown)


Lwmte:
There is actually more recent TRosettaStone edit by E. Popov: http://www.evpopov.com/perso/dl/TRStone/TRLevelFileFormat.html

However, it is much harder to maintain, cause it uses some "smart" formatting...

vvs-:
Shouldn't such docs go to the Wiki? BTW, [openraider](https://github.com/xythobuz/openraider) has it in their Wiki. Don't know about its' age and version, though.

stohrendorf:
I agree that the wiki is logically a better place, but markdown is too limited in semantic (and visual) markup and github's asciidoc is just some markdown flavor. Otherwise I had created a wiki page.

Lwmte:
I also thought about creating wiki for TR documentation, but seems that Github wiki is too simple for this - no categories, no advanced formatting, etc. So I scrapped this idea.

Don't bother with this Popov version, changes are actually not that big, and I already have another TRosettaStone version that I once wrote (it's not on the Net right now). I will integrate both into your remake, plus merge some recent info from TRF (like blending modes, reverb info, specific data types etc.). All in all, we should get comperhensive doc in the end. Also, there are seperate docs about cutseq.pak file format, cdaudio.wad file format, etc., so we should integrate them as well.

PR #170, opened 2015-07-18
move 'updateCurrentBoneFrame' to SSBoneFrame and clean it up a bit

stohrendorf:


vvs-:
Are you working on fixing it properly? Because we need to replace or transform `btQuaternion::slerp` which is incompatible with OpenGL matrices.

stohrendorf:
It's a first step. But i have to study quaternions a bit deeper before I can fully understand the problem and fix it.

vvs-:
I don't think that the problem lies in quaternions per se. There are different implementations which are incompatible with each other. Perhaps Bullets' implementation is perfectly feasible in another coordinate system. I looked at GLM implementation and it's different from both Bullets' and the old OpenTombs' one.

We just must choose something and go with it. The only thing that I know that should work is the old implementation. Actually, I'm looking at reverting the code to it but that would only fix that one problem. To continue working further I'd like to see a clear direction which project developers want to choose. There was a silence on this topic as yet and that stops me in my tracks. I also want to know the reasoning behind replacing the original math implementation with Bullet and why cochrane chose to introduce Matrix4 instead of using the original implementation.

stohrendorf:
I actually don't understand the `slerp` problem fully. As long as both quaternions use the same base system, there shouldn't be *any* differences between implementations, as the slerp operation is mathematically defined, and thus shouldn't differ, except when the quaternions are not unit quaternions.

vvs-:
The formula is the same. It's just that they use different transformations before and after computations. Presumably, because of a different coordinate systems.

vvs-:
Also, there is a new PR which says that the problem lies not in the SLERP itself but in the post transformation in `btQuaternion::setRotation`.

PR #172, opened 2015-07-18
Correct orientation of animations and rotation of objects + various other fixes

stltomb:
What should be fixed:
* rotation of objects
* Lara and other objects are rendered the way they should be

What is not fixed:
* transparency - some rooms are not visible when you look at them through a transparent object
* Lara is leaning to the right, sometimes to a point where her head is upside down

Contains also various other fixes that will hopefully not break any existing functionality.

### DIFF
--- a/src/camera.h
+++ b/src/camera.h
@@ -70,6 +70,8 @@ struct Camera
     Room* m_currentRoom = nullptr;
 
     Camera();
+    Camera(const Camera&) = delete;
+    Camera& operator=(const Camera&) = delete;
 
     void apply();
     void setFovAspect(GLfloat fov, GLfloat aspect);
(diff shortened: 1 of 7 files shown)


vvs-:
Great! Thanks.

Lwmte:
Super! :+1: Going to check it out right now! Seems you've figured out almost everything we were trying to fix up in the past two weeks! :smile: 

Welcome and thank you!

vvs-:
This reveals another round of bugs in triggers, animations etc. But this is a **huge** progress. We could chase these bugs for weeks or even months. You are truly a savior!

Lwmte:
I was kinda aware of trigger bugs, as I've noticed that in my test levels I have wrongly activated entities, which shouldn't be activated. Let's hope they'll be easier to track down.

vvs-:
I believe this PR should be merged ASAP. It fixes a vast amount of problems.

Lwmte:
Ok then! :)

PR #198, opened 2015-07-21
Updated the documentation with SCRIPT.DAT specification

ActiveMath:
Included SCRIPT.DAT/ENGLISH.DAT specification as used in TR4 and official TRLE. As far as I know, TR5 uses very similar chunk IDs, with the only difference being the weather command and possibly some command that has to do with Lara losing her weapons in one of the Russian levels. Need to find this out.


stohrendorf:
This is markdown, not asciidoc. Please keep the style consistent, and rewrite it like the rest of the document.

Lwmte:
It's OK, I will reformat it! Thank you very much! :+1:

PR #202, opened 2015-07-23
Rendering fixes

stltomb:
Fixes:
- Lara's rolling animation (tested in TR1)
- Lara's position in noclip mode
- animations of entities (bats, doors, levers, ...)
- inventory and item notifier
- position of the zip-line handle (tested in TR2)

All patches are fairly simple and were made just by comparing the code before the C++ merge and after it.


Lwmte:
Thank you! You've saved us a lot of time!
I think we should add you to the contributors list...

stltomb:
Thanks.

PR #213, opened 2015-07-24
Replace matrix4 with glm::mat4

stohrendorf:
GLM is header-only, widely available, and we don't need to re-invent the wheel.


cochrane:
If you insist, I don't mind. But I'm still not sure what exact problem you're trying to solve by replacing existing code with new code, or how that is anything but reinventing the wheel.

stohrendorf:
I don't insist, this is rather some sort of "proof of concept"; the benefit I see is that GLM is much better integrated with OpenGL than bullet, as the internal data layout is the same as OpenGL expects. Also, it provides some nice extensions that would reduce `vmath.*` to some sort of bullet/TR/GLM conversion lib.

Lwmte:
Do we really need it right now? I think it's better to hold back any introductions of new libraries, until it is critical and absolutely needed.

From what I've read, glm is poorly documented and also have issues with angles vs. radians.

PR #220, opened 2015-07-24
move sound related members from World to new struct AudioManager

stohrendorf:
At least on windows, I often get an out-of-memory message from OpenAL, and we generally have some crippling noise and also rare memory corruptions. This is a first approach for decoupling the audio interface from the engine.


Lwmte:
Getting quick crash with this one, when Lara stops running.

stohrendorf:
I'll close this pull request, as it is too risky to merge.

PR #224, opened 2015-07-25
Animation control cleanup

stohrendorf:
CXXify ClimbInfo and split up that monstrous state_control_lara into subfunctions and move them to the proper place. I have tested it, and it seems *not* to introduce new bugs. Still, the old ones are there.


vvs-:
Yes, seems that no new bugs are here :smile:

vvs-:
Sorry, there is new bug here :disappointed: Try jumping to the left.

stohrendorf:
Forgot to add the "jump right" case label when merging the left/right state control functions.

Lwmte:
Oh, I dreamed about splitting up this monstrous case switch into several subfunctions for a loooong time... :smile: 

But is it really necessary to move all anim_state_control code into character_controller? I mean, it seems logical, but code gets too "condensed" into several files. We already have this issue with ragdoll/hair, I had these separate modules especially for the purpose of getting code a bit less chaotic, but now these files contain almost no code.

Lwmte:
I propose to remove case switch completely and replace it with function pointer table.

stohrendorf:
I just wanted to make sure that this actually works before removing the switch and maybe introducing another bug.

Speaking about "modularization": Yes, animation/state control should be outside of the character, but this will only work when every data member the controllers accesses is encapsulated in getters/setters/convenience functions, and the data members itself are made private, so that you can better track which state functions actually change what data member.

Lwmte:
You mean all these HeightInfo, low_vertical_space, last_frame fields? Yes, I agree with you. Especially regarding HeightInfo - this monster should be destroyed some day (and by the way, it is current engine bottleneck).

So far everything works, at least I haven't seen any bugs.

Lwmte:
P.S. ClimbInfo is even more monstrous than HeightInfo.

Lwmte:
So, how everything's going?

stohrendorf:
Haven't worked on it anymore, because I wanted to have the master branch crash-free first... now I know that it never really was.

Gh0stblade:
Yes, I believe that we need to keep that implementation of "look at" for the camera. It's not just fixed cameras where this is required. Fixed cameras are fairly easy to implement you just either create another camera then switch the renderer's current camera to the fixed one. Or have the current camera's position overridden. Sometimes a TARGET is required for fixed cameras and usually the camera must look at a specific entity in the world. 

LookAt functionality is also required for cutscenes, cinematic frames usually have a target X, Y, Z (according to TRN)  to point the camera at. When the aiming system for Lara is implemented, we can just set the entity's position as the target to keep the camera facing the enemy.

stohrendorf:
Closing this, as it's too risky to merge it.

PR #235, opened 2015-08-02
VS2015 Buildability

stohrendorf:
This finally allows to create a VS2015 solution and make OpenTomb buildable and runnable in it.


Gh0stblade:
Great work, this is something I've always wanted to become possible! I approve.

vvs-:
It doesn't link on Linux, at least SDL2 detection is broken.

stohrendorf:
Can't confirm this. It detects SDL2, compiles, links, and runs here.

vvs-:
But it doesn't link here. Fedora 21 32-bit.

I bet it'd quickly become broken after merge because nobody will test it.

stohrendorf:
Where is your SDL2 library located and where are the headers? And what exactly does "doesn't link" mean, i.e. are there missing symbols, or is the library not found? Maybe the find script needs some extension.

vvs-:
Fixed by removing `SDL2MAIN_LIBRARY` from `CMakeLists.txt`.

stohrendorf:
That doesn't answer my question.

vvs-:
Well, it doesn't find SDL2MAIN_LIBRARY, so it can't link it. There is no such library in Fedora.

stohrendorf:
OK, that brings me to a solution, because especially on windows, sdl2main *is* needed, yet on other platforms, it is optional, but often enough it is present as a stub library. I'll see if I can make it optional.

stohrendorf:
Turns out it was already handled by the FindSDL2.cmake.

vvs-:
It  doesn't work anyway. I get `inf` instead of the FPS indicator. I have no logo or splash screens. The level maps are not loading for some reason and there are no any error messages.

stohrendorf:
I guess I found the problem, but I can't test it right now. Probably MSVC treats `1e3` as float, while GCC treats it as `int`, which caused a wrong `Sys_FloatTime` result.

vvs-:
While `setgamef` doesn't work I'm able to load maps by name. But everything is frozen.

stohrendorf:
Does it show somewhat correct FPS information?

vvs-:
No, it's still `inf`. It briefly shows 200 fps at the beginning.

stohrendorf:
This is the last try to fix it here from windows. When I can boot to linux again, I'll try to fix it there if this doesn't solve the problem.

vvs-:
Seems to be working again. But I didn't thoroughly test it.

vvs-:
I see that time is now in milliseconds while `gettimeofday` returns time with a microsecond precision.

vvs-:
Now everything looks like in `master`.

PR #247, opened 2015-08-04
use in-tree built bullet

stohrendorf:
builds bullet in-tree and forces the usage of the in-tree build bullet version


cochrane:
Looks good to me. Initial test confirms it works well for bullet. If this gets accepted, I'll probably try and use that approach for other libraries we rely on as well.

stohrendorf:
I had a thought about it, and maybe it's even better to put the bootstrapping parts into a separate repository, used only for preparing the build environment. This way we would keep the main tree free from external code at all, and that "build env setup" could get more elaborate, e.g. by setting up also SDL2, sndfile, etc. for users of MSVC.

cochrane:
Yeah, that might be a good idea. Right now this works, but I can't directly generate an Xcode project, and I've found cmake's ability to create projects for use by IDEs very useful.

stohrendorf:
What is the problem when trying to create an XCode project? Is it this approach in general, or something else?

cochrane:
I get an error that no Xcode project can be generated for bullet. Not sure why, or why I should care (since I don't need one for bullet anyway). Maybe it is possible to generate one now that bullet has been built. I may need to do more experimentation when I'm back from work.

stohrendorf:
This is because the bootstrapping inherits the project type. It can only be cleanly solved (without overwriting/hacking the cmake generation and make programs for bullet and openal) by moving the bootstrapping out of the main project.

vvs-:
It does work here. I didn't notice any problems.

cochrane:
> This is because the bootstrapping inherits the project type. It can only be cleanly solved (without overwriting/hacking the cmake generation and make programs for bullet and openal) by moving the bootstrapping out of the main project.

That does sound reasonable.

stohrendorf:
This is deprecated, see #254.

PR #251, opened 2015-08-05
replaced bullet's unstable quaternion slerp

T4Larson:
Bullet's builtin quaternion slerp may produce result quaternions full of NaNs when the interpolated quats are very similar or identical.
This happens very frequently in an optimized build of the engine (gcc 4.9.2 with -O3), where Lara's limbs flicker randomly, causing mid-walk stumbles and on some maps (karnak.tr4) produce and "overflow in AABB" a couple of seconds after map start.
With the patch, an optimized build seems to work fine so far...

### DIFF
--- a/src/entity.cpp
+++ b/src/entity.cpp
@@ -770,7 +770,7 @@ void Entity::updateCurrentBoneFrame(SSBoneFrame *bf, const btTransform* etr)
         if(k == 0)
         {
             btag->transform.getOrigin() += bf->pos;
-            btag->qrotate = src_btag->qrotate.slerp(next_btag->qrotate, bf->animations.lerp);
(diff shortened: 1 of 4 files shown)


Lwmte:
Well, hello there! :smile: It is so nice to see you again! I saw that you forked engine some time ago, and immediately thought that something cool is coming! :sunglasses: 

I will test the patch right away. It will be also a good idea to report this problem to E. Coumans as well (https://github.com/bulletphysics/).

Hope to see more contributions from you! :+1:

Lwmte:
Seems that everything is fine, I'll merge it right away.

T4Larson:
Thnx, I've just sent a corresponding pullreq to bullet: It seems there is an older reported issue (bulletphysics/bullet3#64) with that issue from last year.
We'll see when/if thats merged in bullet - I'll try to remember to switch it back when/if it's fixed in a release ;)

Issue #28, opened 2015-05-14
Fallback state implementation and unused speeds


Lwmte:
Copied from Tomb Raider forums (post by Joey79100):

> When there's a pit behind you, and you press Backwards, Lara will start hoping backwards. Once she's above the empty, she starts falling, playing animation 34 and then 35 (that have StateID 3, that is used for jumping forwards), but it's wrong. She should play animation 93 that has StateID 29 (and without animation 34 before).  
> This animation and its StateID were considered as unused by Fluen when creating his animation list (and consequently, StateID29 wasn't referenced in the StateID list), because visually it looks the same as animation 35, but actually it is used, and there's a reason.  
> The problem that exists now is that if you are already pressing the Action key when jumping backwards (to grab the ledge), Lara gets stuck on the edge, switching continuously between animation 22 (starts falling while jumping forward when she hasn't finished the "jump cycle" (when there was a ceiling)), and animation 82 (falls on the ground).  
> Animation 93 lets you grab a ledge only after 7 frames (not a hardcoded thing, the StateID change is written this way in the animation), while animation 35 lets you do this during the whole animation. So using animation 93 will prevent her from being stuck like she currently does.
> 
> Proof that the StateID exists: in TOMB.MAP that was leaked with the May TR1's demo, it's referenced this way:

`0001:00013da0  lara_as_fallback_`

> (StateID 30 is used when shimmying left, and is just after 29 which has a rather clear name)  
> I don't know if you've written code for this StateID though, but there shouldn't be a lot of differences.  
> By the way it seems that the speed for animation 95 (ready to grab) is ignored. She's meant to go forward, but in OpenTomb she has no horizontal speed.  
> The speed she has when she hits a wall while jumping are ignored too.

Joey79100:
Much better now, but animations 34/35 are still played when after sliding backwards, and it causes Lara to make... weird things.

Lwmte:
Speeds are not fixed, though.

Gh0stblade:
The speeds are completely broken, I've been comparing things to the original Tomb Raider engine in an attempt to fix the problems. I have an almost *i think* working solution but it breaks on TR3 sprinting and for some reason the speeds are still not being applied correctly (on the roll state) it has to be something deeper than just issues with calculating the speed/accel values.

I've also noticed Lara's speed is not updated quickly enough after slowing the game down? Could be wrong though. I'll try fix this.

Gh0stblade:
[...]

PR #255, opened 2015-08-08
Add acell limits like original

Gh0stblade:


Lwmte:
Seems like sprint and run speeds are now broken. Lara's acceleration is too low.

Can you provide some examples where these fixes are relevant?

Gh0stblade:
Hi Lwmte! I'm not done just yet. I think i've found something very important related to the speed/acceleration system. If it's confirmed I'll add some more commits with an explanation here so that it can be discussed further.

Cheers.

Lwmte:
Ok, I'll keep an eye on it. Thanks! 
Do you think you can manage integrating speed calculations into freeFalling move type?

Lwmte:
Offtopic: @Gh0stblade - please take a look at #256 :wink:

Gh0stblade:
You mean vertical speed? I believe it is hard coded. I'm currently looking at TR5 and debugging it to see what exactly is going on and why these speed issues are occuring. I'm not sure if it's worth it though since the system is completely different to OpenTomb and probably wouldn't be compatible but if we get the math right, it should still work.

Anyhow, this pull request is being closed. There will be another one.

Lwmte:
No, not vertical speed. I mean that during freeFalling move type, animation speed should be processed as well. If you compare, for example, moveOnFloor code with freeFalling code, you will note that freeFalling is missing that four-part if-then statement ([look at line 1008](https://github.com/opentomb/OpenTomb/blob/master/src/character_controller.cpp#L1008) in character_controller.cpp).

This is the ultimate reason why horizontal speeds are not processed with free falling states (mentioned by @Joey79100 in #28):

> By the way it seems that the speed for animation 95 (ready to grab) is ignored. She's meant to go forward, but in OpenTomb she has no horizontal speed.
> The speed she has when she hits a wall while jumping are ignored too.

Lwmte:
Currently, wall hit animations and animation 95 speeds are done using ugly workarounds.

Gh0stblade:
@Lwmte are you sure? I see the speeds correctly being applied.

Issue #38, opened 2015-05-26
Lara jumps too far


vvs-:
In *Caves* when Lara prepares to make a standing forward jump to the last secret from the ledge, she can only grab the opposite ledge to climb. But in OpenTomb she safely land behind the gap.

Lwmte:
I think it is related to jump speed issue - in OpenTomb, Lara jumps significantly slower than in original. I will work on it after I finish hair and ragdoll functionality.

vvs-:
`setEntityPos(player,77897,33218,-3072,83,0,0)` in *Caves*, then jump forward. Lara bumps her head against the rock. This doesn't happen in the original engine, i.e. she jumps too high.

Joey79100:
It looks like many speeds are a bit "too much". If you're on the edge a block and you press backward, normally she should make a little hop and reach the opposite edge of the block, but instead her hop is a bit too long and she falls off it.
Same, when you have 2 blocks separated by a 2-block long pit, you can easily jump from one to another just by jumping from the edge of the first, while in the original games you need to press Action to grab the ledge, else you fall.

vvs-:
Actually it's not just the constant. Modifying default speed makes Lara jump on a too short distance in one cases and still too far in others. Something is wrong in the whole formula.

Lwmte:
Yes, it's completely wrong. Currently it is linear formula, i.e. each frame Lara is pushed at a constant height upwards, until affected by gravity. In original, jumps used either logarithmic or trigonometric formula, so Lara got larger momentum at the beginning of jump, and lower momentum at the end.

stohrendorf:
A better approach would then be `speed[n+1] = speed[n] * (1-alpha*frameTime)`, where `alpha < 1` describes the damping/downwards accelleration per second.

vvs-:
Can't the original computation be reverse engineered?

Lwmte:
Yes. I tried to do it, and here is my result. Here's a snippet of jump code from TR1:

```c
  if (lara_pointer + 68) & 0x08 )  // 8 is a JUMP STATE flag.
  {
    lara_pointer_ = lara_pointer;
    
    // Some tricky business happens here...

    v70 = v67 + v68 * (*(_WORD *)(lara_pointer + 22) - 1 - *(_WORD *)(v81 + 16));
[...]

PR #258, opened 2015-08-09
Correct several issues regarding incorrect calculation of speed/acell

Gh0stblade:
See #28 #38 

This should resolve some bugs. Currently performing some tests to see how close it is to the original game (the speed_mult side of it and gravity)


Gh0stblade:
Yep, I decompiled the code from TR5 for it. Speed/Accel values per frame are 100% like the original now and it's important to find out where these speed issues are coming from. There's another bad bug that's causing Lara to move too far when she rolls but the accel/speed from animation is correct like the original. I think it's bullet related.

There are many data type issues tbh. For example, the original game used signed integers where OpenTomb uses floats. I'm not sure what the lateral speed/accel was for TOMB4/5, I'm guessing they should just be integers since "mixfloats" don't actually exist but as I've not seen the data I can't confirm whether it's the same as speed/accel and thus have left it.

As for the speed_mult, I will take a look at TOMB5 to see what's going on there.

Cheers.

Lwmte:
There's no difference between normal and "lateral" speed/accel, they are basically the same types of values, with latter being used *only* for non-Lara entities. "Mixfloat" (or fixed-point, as properly called by @stohrendorf) is actual data type used there, just in TR it is calculated in runtime, while in OpenTomb I tried to make it more like in WadMerger, converting them to float at loading-time. I still think it's better to convert them to float at loading-time, since OpenTomb is float-based, and there is no point using bitshifting operations just to get the speed right.

vvs-:
It may be closer to the original, but I don't see the difference. In some cases it performs even worse than before. On the other hand there could be other problems with speed/acceleration which are not addressed by this pull request. We need to find them to get a fully working solution.

Gh0stblade:
@vvs- Yes, that's correct. To clarify, this is a patch for *only* the x speed of Lara which is obtained from the animation. @Lwmte I think it's best we stick with this solution. The currentSpeed is still a float, it will just be whole numbers and this is the "magic". If we want to get it working properly, we have to start from using the correct formulas from the original game. Also, I believe you're correct about the 30.0f time-step. This will be reverted.

@vvs- There are some serious problems with the speed calculations else-where. But this produces animation speed/accel values 100% the same as the original games and will help us debug Lara's problems more easily. There might not be any noticeable changes, but if you view it frame by frame, log the speed then compare it to any TR game it's the same but before it was not and being precise is extremely important.

Cheers.

vvs-:
It's better now. But Lara still can't reach one ledge in *Atlantis*. BTW, I like this debug info.

Gh0stblade:
@vvs-  Animation 95 has no horizontal speed at all according to the animation data. You can also check this in TRViewer.

vvs-:
That one: `68154,45984,-10240,0,0,0`.

Also, I see speed 48 during the animation 95, while there is no horizontal speed in animation 77. Though, I doubt that there is any difference in the distance unless there is a ledge.

Gh0stblade:
Animation 95 should have a speed of 8 so that's ok.
As for Anim77 that's tricky. The original TR5 somehow gets the speed_x value of 50. The actual animation doesn't have horizontal speed set. Maybe it was exclusively implemented for TR5 or something, could also be hardcoded.

Lwmte:
Hmm, but that effectively makes current formula the same as I used in the first place! :cool:

Lwmte:
By the way, I know why Lara is misaligned on twin roll by the wall. It's again caused by "fixing" body part positions. I'm going to fix this in a minute.

Gh0stblade:
Great find @Lwmte  Now all we need are some tweaks to the collision so Lara's position is the same as the original when she hits walls. I believe she's not close enough to the walls so sometimes going back or rolling on a solid square surface results in Lara falling off.

PR #262, opened 2015-08-10
Vt loader refactoring

stohrendorf:
TR 1, 3 and partially 4 (the included test files) work. The other games need a few tests if they load successfully. If anybody finds a fix, please commit.


Lwmte:
Nice! :+1: I will test it right away.

Lwmte:
I can't compile it, CMake says this:

```
CMake Error at src/loader/CMakeLists.txt:1 (add_library):
  Cannot find source file:

    tr1level.h

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx


CMake Error: Cannot determine link language for target "vtloader".
CMake Error: CMake can not determine linker language for target: vtloader
```

Lwmte:
Ok, I have removed VT-related stuff from CMakeLists.txt, and now I get these errors during compilation:

```
undefined reference to `loader::Level::createLoader(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, loader::Game)'

undefined reference to `loader::Level::createLoader(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, loader::Game)'
```

Lwmte:
It seems that we're missing several header files here, from `tr1level.h` to `tr5level.h`.

stohrendorf:
Forgot to add them, sorry.

Lwmte:
Ok, now it doesn't complains of anything, but compilation still fails. I just get thrown to Makefile by Code::Blocks with error message `recipe for target 'OpenTomb/fast' failed`.

Is CMake script broken in this pull request? Should I try again removing vtloader from CMake script and instead point to it manually in PRJ_SRC/PRJ_HDR?

Lwmte:
Yes, seems that CMake script is broken. I've removed /loader/ script inclusion and instead made it a part of the source, and it compiled successfully.

So far TR1-4 levels are loaded successfully, but I have crash with TR5 levels... Not actually a crash, but possibly incorrect parsing. The log from command window:

```
Bad value for flags[1]
Bad value for flags[2]
Bad value for flags[3]
Bad value for flags[4]
Bad value for flags[5]
Bad value for flags[6]
Bad value for flags[7]
Bad value for 'unused'
terminate called after throwing an instance of 'std::length_error'
    what(): vector::reserve
```

src/loader/tr5level.cpp
```diff
@@ -0,0 +1,293 @@
...
+       0 No weather
+       1 Rain
+       2 Snow (in title.trc these are red triangles falling from the sky).
+     */
+#if 1
+    m_reader.skip(2*sizeof(uint16_t));
+#else
+    i = m_src.readU16();
+    i = m_src.readU16();
+#endif
```
Lwmte: Can you please integrate this into TR_Level structure? Seems these two values are ignored. You can create global flags for level structure, one called `m_LaraType` and another called `m_WeatherType`.

Actually, `m_WeatherType` can be also re-used for TR3, where weather was hardcoded.

Lwmte:
Still same error...

stohrendorf:
Unfortunately, I can't solve this anytime soon, because I can't find my TR5 CD anymore to test it locally. But I guess either there are too many/few textures loaded, or I missed some data type differences.

Gh0stblade:
You can download the PC demo.

Lwmte:
Or just download any TR5 custom level. For example, [this one](http://trle.net/sc/levelfeatures.php?lid=1904).

stohrendorf:
I tried to debug it, but found no obvious errors. Until that exception gets thrown, everything looks fine and sane. In that custom level (`joby3.trc`), there are 3 lights, but the second `numLights` value is 0, so the loader throws an exception. Did the "old" engine load these levels? Because I can't find any data size loading differences in the loader code up to that specific point.

vvs-:
No, this level doesn't load by any version of OpenTomb.

stohrendorf:
Then it seems that these two `NumRoomLights` numbers aren't always the same.

Lwmte:
Yeah, seems to be bad example, as level was created by Dxtre3d, which is built on the same info we had prior to TRosettaStone! :laughing:

Here is [first level from retail TR5](http://trep.trlevel.de/temp/andrea1.7z), I hope nobody sues me for publishing solely this file? :space_invader:

Lwmte:
And yes, most likely, old engine successfully loads this level (if it got release on trle.net at all).

It's just VT loader being **very** strict on filler/unused values, it just crashes an engine in case it sees something unusual. It's very possible that some "unknown" or "unused" field (as `NumRoomLights`) was treated differently in Dxtre3d for some reason.

Unfortunately, as TPascal decided to keep his editor closed-source, only he can know why he did that.

vvs-:
> I hope nobody sues me for publishing solely this file

I doubt that anyone could sue you for that, but GitHub could get in trouble. And they in turn could create trouble for this GitHub organization.

Lwmte:
Ok, I'll remove the link then! :smile:

stohrendorf:
@Lwmte That works fine here, as long as it is expected that in that level Lara has "black zones" (where she is fully black) and "textured zones", and that the connection between her head and her hair is missing.

stohrendorf:
I can't see any bugs related to loading on my side anymore, so if you all have no objections, I'd like to merge this.

Lwmte:
Seems that CMakeLists.txt is still broken... I can't generate working CodeBlocks project with it.

stohrendorf:
What exactly is the error?

vvs-:
*The Great Wall* doesn't load:
```
terminate called after throwing an instance of 'lua::RuntimeError'
  what():  scripts/entity/entity_model_ID_override.lua:166: attempt to index global 'tr4_id_override' (a nil value)
Aborted (core dumped)
```

stohrendorf:
That's shouldn't work with current master, too, as @Lwmte commented some tables out in 30a225035c24fd5e412bc0ac45799a256518c3a2

vvs-:
Actually, it does work in `master`.

stohrendorf:
Then I'll take a look at it.

stohrendorf:
Hm... actually, it shouldn't work in master, too, because `tr4_id_override` isn't defined, so any ID override for TR4 and TR5 should fail.

vvs-:
That's how it looks in `maste`:
```
        if(tr4_id_override[id] == nil) then
            return -1;
        else
            return tr4_id_override[id];
        end
```

vvs-:
Actually, why there should be tr4 override for TR2? It looks like a wrong version.

stohrendorf:
Sorry, my fault. I forgot to convert the game ID to the engine ID when passing the argument to the script.

vvs-:
> That works fine here, as long as it is expected that in that level Lara has "black zones" (where she is fully black)

That means that TR5 is broken because it works differently in `master`.

Lwmte:
Why? We had green Lara meshes in catsuit levels all the time.

On the recent changes: now I can't even generate project with CMake because it complains of missing resources/icon/opentomb.rc file. I guess it happened because CMakeLists.txt was moved to /src folder.

vvs-:
I'm talking about black Lara model at the start of *Streets of Rome* while it works in `master`.

stohrendorf:
Seems like in rooms with dynamic entities, every dynamic entity is either fully unlit or fully lit, and sometimes Lara seems to be correctly lit. I wasn't able to find the cause yet, but at least I know where to search for it.

vvs-:
I can just say that point lights in TR5 was broken all along, but the ambient light should be ok. And seems that the ambient light is now broken in that branch.

stohrendorf:
Ambient light is probably currently also broken in master, because the light intensity wasn't initialized. I guess it is pure luck that the float is initialized to an arbitrary value that gets clamped to 1.0.

vvs-:
I didn't notice any more bugs.

stohrendorf:
When @Lwmte and @Gh0stblade give their OK too, I'll merge this.

Lwmte:
I still can't compile it with CodeBlocks. It throws me with "Recipe for target failed" error to `Makefile` on this line:

`124: 	cd /d C:\DEV\projects\OpenTomb && $(MAKE) -f src\CMakeFiles\OpenTomb.dir\build.make src/CMakeFiles/OpenTomb.dir/build`

stohrendorf:
@Lwmte Can you give me the full log, please? I don't have any idea what is wrong.

Lwmte:
There's not much to see in full log, it just throws some warnings about unused parameters, and right after this it throws that single error without any extra info.

I believe something is wrong with Makefile generated by CMake. If I remove `add_subdirectory(loader)` line from CMakeLists.txt, and instead make loader as part of the source code (just including its paths in PRJ_SRC and PRJ_HDR), it compiles normally, although working directory in CodeBlocks project is set to /src instead of root folder, so I can't directly launch the engine from CodeBlocks - instead I need to move OpenTomb.exe from /src folder back to root folder and launch it from there manually.

Lwmte:
As for the loader itself - yes, seems that everything works perfectly! Great job! I even see that you've fixed TR5 lights? :+1: 

P.S.: So far never encountered "black Lara" bug in any TR5 levels.

stohrendorf:
That's very strange, but I can't see any error in the CMake files, except that I am a bit concerned about that MinGW hack.

Lwmte:
I believe it's because something is wrong with static linking of loader library. If I compare Makefile generated without adding loader as a library, same line which previously resulted in a error (see above) looks like this:

`122: $(MAKE) -f src\CMakeFiles\OpenTomb.dir\build.make src/CMakeFiles/OpenTomb.dir/build`

As you can see, there is no `cd` command. Maybe that's the reason?

Also, seems that after creating one extra CMakeLists.txt in /src folder, CMake sets working directory for a project in /src too, which results in binary misplacement bug I described above.

stohrendorf:
Can you try to make a clean out-of-source build?

Lwmte:
I'm doing clean build every time I switch branch, clearing CMake cache first, then reconfiguring it and recreating CodeBlocks project file, and then pushing "Rebuild" button in CodeBlocks (which is equal to `make clean` command).

Or I need to make some more additional steps?

Lwmte:
If this issue is not going to be fixed anytime soon, can we just include loader as usual source code part, *not* making it external library for now? It can be made later.

stohrendorf:
With "clean" I mean "clean like a laboratory". This means a completely independent, fresh checkout, and an out-of-source build (`mkdir build && cd build && cmake ..` so to speak). I have the suspicion that there are still some CMake things left in your directory.

An yes, it can be made part of the previous build layout again if the above doesn't work. But I leave that to you, because you seem to got it working.

Lwmte:
I have completely removed all "CMake*" files and folders from everywhere in my project folder, as well as all Makefiles and such, leaving only original source files. Then I have rebuilt CodeBlocks project in CMake - and still the same result.

So I suggest one of two ways to deal with such situation: first is making loader a part of OpenTomb for now, leaving the task of making it separate for the future. And second way is moving loader out of OpenTomb to its own repository, and then refering to it as a submodule, in the same way as LuaState, Bullet and OpenAL.

stohrendorf:
OK, let's do the first approach and keep it inside OpenTomb, as it is still coupled to it. Can you commit your changes to the build system so it can be merged without breaking anything?

Issue #266, opened 2015-08-11
Speed issues


Lwmte:
I have noticed that after my recent changes with collision shapes and flags, FPS has significantly dropped, especially when Lara is in spikes, and even in death state. This is certainly not Lua issue, as collision callback function never enters if Lara has no HP, so most likely it's Bullet slowing down with bodies without contact response?

vvs-:
Yes, I noticed it too. I have a significant drop in frame rate at the beginning of some TR5 levels. If I do nothing and just wait then it improves after some time.

vvs-:
Seems that sliding is very expensive now. Indeed, it looks like a Bullet issue.

vvs-:
On the second thought, I have no slowdown on the same slope after some time. Seems that there is something at the beginning of the level.

vvs-:
It was broken in a092d3d38272b68c765778fc97141636296e392b.

Lwmte:
I guess it's one of the collision flags which produces this bug. Possibly, CF_CHARACTER_OBJECT is too expensive to be applied to every NPC (however, all NPCs are disabled until triggered, so you surely shouldn't get any slowdowns on level start-up).

vvs-:
Try *The Base*. When Lara start to slide she's awfully slow. But after the grating falls everything returns to normal.

vvs-:
This seems to be exactly the time required to pre-activate all entities. So, this slowdown happens during level initialization.

vvs-:
I've noticed that these entities have ghost collision mesh at the beginning but turned to dynamic collision mesh after initialization. I've tried to look at it in the process using `r_coll` and voila, I've got a crash:
```
#0  0x08360f93 in btDiscreteDynamicsWorld::calculateSimulationIslands() ()
#1  0x08366aa4 in btDiscreteDynamicsWorld::internalSingleStepSimulation(float) ()
#2  0x0836037f in btDiscreteDynamicsWorld::stepSimulation(float, int, float) ()
#3  0x08303237 in Game_Frame (time=0.0543360002) at src/game.cpp:842
#4  0x08229666 in Engine_Frame (time=0.0543360002) at src/engine.cpp:522
#5  0x08336c6a in main () at src/main_SDL.cpp:26
```

vvs-:
And it wasn't ghost before a092d3d38272b68c765778fc97141636296e392b.

vvs-:
[...]

Issue #261, opened 2015-08-10
Animation bug roll/direction bugs.


Gh0stblade:
I've been experiencing some animation bugs that occur when Lara rolls in either mid-air jump or on floor.

The bug in question results in Lara's rotation being somehow inverted so she's facing the opposite direction she should be. This usually occurs at low framerates.

vvs-:
If a framerate is significantly lower than 60 then animations start to break down. This is because the original engine used fixed framerate and OpenTomb need to simulate that. Using conversions between integer and floating point numbers result in rounding errors and uneven distribution of physical frames between simulated ones. Sometimes simulated frames and/or state changes get lost in process. I did some specific fix to compensate for that in a particular bug, i.e I put the same simulated frame twice in a single physical one. But this isn't a panacea and it still could happen in different circumstances and a very low framerates.

Gh0stblade:
@vvs-  Yes I agree, original games used whole integers. The float issues are also the reason for most of the speed/position bugs even though one was fixed. As @T4Larson also suggested OpenTomb should simulate physics animations at 30 fps, it's the only way these issues due to floating point rounding errors will be solved.

Lwmte:
I don't think there is a point in going back to 30 FPS. One of major reasons for developing OpenTomb was to get rid of 30 FPS limit once and for all. Physics and scripts are done with 60 FPS speed, which is not so different from 30 FPS timestep - you usually can divide everything by 2 to make it genuine.

I can only propose "decoupling" of animation engine and actual speed/acceleration calculations, which is halfway done by updating speed/accel only on frame changes. So we now need to move all these `Character::freeFalling`, `Character::moveOnFloor` etc. to fixed 60 FPS timestep. And I really mean 60 FPS timestep, cause having three different timebases in one engine is already too much.

Along with speed/animation, Bullet rigid bodies must be also updated on a fixed timestep, i. e. there should be no calls to `updateRigidBody` and `updateGhosts` in variable timestep loop. Basically, *only* thing which should be done on a variable timestep base is updating camera position, animations and graphical effects (animated textures, etc.).

So that's how I see it:

* Decouple speed and acceleration calculations from variable timestep (already done, but needs to be moved out of `frameImpl` function somewhere else, out of animation code).
* Decouple rigid body and ghost object updates from animations.
* Decouple collision checks from animations.

So the only code which should be executed *every* frame is animations themselves.

[...]

PR #269, opened 2015-08-12
converting to fixed physics timesteps

T4Larson:
With the current timestep system come some issues:
* low fps problems:
    * logic actions and animcommands may be skipped
* high fps problems:
    * physics functions called too often and from too many submethods (redundancy)
    * acummulating precision errors may become apparent 

As I found out so far, this is a bit like stirring up a hornet's nest, a lot of entanglement between object physics movement and game logic parts, there's still a bit more to be done:
* Animation transform updates and game logic effects should be separated,  to allow interpolation without unneeded logic and physics updates, and better maintainability
* reduce usage of global engine_frame_time in functions and methods that must be called in multiple substeps (low fps issue)
* move game logic actions to internaltick or use updateAction from btActionInterface: This removes the need for an extra catch-up loop for the gameticks on low fps because bullet does this already in the substeps...
* interp animations every frame, i.e. only update interpolated bodies for the renderer, but don't do physics - this needs some disentangling in the code...

This isn't ready to merge yet - although it works and fixes at least the skipped-animframes, I'm currently still figuring out some frame/logic depedencies...


vvs-:
I'm curious in how well it'll turn out in the end. I've already seen how badly entangled those steps are and how it affects the core of the engine.

Lwmte:
This is really cool that you have decided to fix system-critical bugs in timing. Seems that TeslaRus is not going to make any major changes here anytime soon anyway...

vvs-:
TeslaRus makes fixes in his branch. Unfortunately, it's diverged and became a fork. Now to get benefit from his fixes somebody need to port his code to a different code base and that's a lot of work.

Gh0stblade:
I tried this last night and the animations were pretty buggy, saw no differences except Lara's speed is much faster but it's not finished right now.

Pretty cool to see Tesla still updating his fork, only problem is it's very outdated compared to the master and thus merging such changes will require effort and furthermore might cause bugs.

vvs-:
> saw no differences except Lara's speed is much faster but it's not finished right now.

Speed isn't constant, but indeed it's incomplete yet. I think that the described design is sound and is a move in a right direction.

T4Larson:
> I tried this last night and the animations were pretty buggy, saw no differences except Lara's speed is much faster but it's not finished right now.

I haven't seen any animation bugs yet, but it's true: speed is not only faster, it also varies slightly when you run around - the timing decoupling isn't finished yet, and my suspicion is that the the entity movement methods don't handle very small times very well (the remainder time afte the catch-up loop) and cause these irrgularities, especially with vsync at 60fps enabled, where the remainder time is only the slight difference of the screen refresh from 'real' 60hz based on the internal clock.

Other things may show up that are bugged elsewhere, i.e. with low fraterate, turning l/r may spin wildly when reaching max. turnspeed, but hopefully having a predictable timing will make the overall movement behaviour more consistent.

Gh0stblade:
Are you sure? Lara is completely deformed, it's like some of her body parts are incorrectly rotated? This happens with some animated models as well.

vvs-:
I really don't like these inconsistencies which can't be reproduced reliably. First it was Lwmte with sliding and rotation bugs, then Steffen with falling floor and now you have another one with Lara animation. Something fishy is going on. I don't know where to look: either some memory corruption or something even worse.

T4Larson:
> Are you sure? Lara is completely deformed, it's like some of her body parts are incorrectly rotated?

:confused: Ugh... I see none of this, other than speed everything seems to be alright. What's your framerate?

Gh0stblade:
@T4Larson 

The issue happened at both high and low framerates. Right now, there are no problems at both low and high framerates either. No idea what caused this but it seems to have gone away *for now* it's probably not your code that's causing it but something else as @vvs-  Suggested there are some weird inconsistent bugs that rarely happen.

vvs-:
Perhaps, I have something similar too. I've noticed that sometimes Lara can't jump high enough to grab the ladder. And after some time it disappears.

T4Larson:
Yeah, I see the organic growth... 
Currently it's like poking a hornet's nest while running through a chain reaction, fixing one thing causes errors elsewhere to become apparent, while the workarounds swarm around and bite you in the back... :stuck_out_tongue_winking_eye:

stohrendorf:
This looks promising. What's your progress?

T4Larson:
I think I'm close  to where I think I can stop fixing code affected by the conversion, only major thing left to do is fix the weapon handling.

What probably remains unfixed is a issue with the ragdoll: Joint stiffness seems to vary with simulation rate (this appears to be because of how bullet handles joint-ERP per simstep?) - this effect is already noticeable with the current master, where the ragdoll appears more stiff without VSYNC (higher rates).
In this version she collapses more like a blob, but I think that can be addressed separately.

T4Larson:
> Shouldn't this be implicitly always true, as Lara is the only instance of Character that get assigned to engine_world.character?

The existing code handles Character objects in the world.entity_tree (see Game_UpdateCharactersTree()) which may not be the player-Lara - with slight differences - and this is preserved through the isPlayer() condition in the frame() Method of all Characters.

It may not be used right now, but it's nothing that is clearly made obsolete by the framestepping changes, so I'd better preserve this behaviour.

The original idea behind that probably was to have NPCs based on the Character class which are controlled through the same command-action interface (allowing nifty features like a player taking control of an NPC or having a ghost-Lara run along with you on a timerun with a stored command sequence...), maybe @Lwmte can tell us more.

The idea is still something to keep in mind, imo - although the Character class has developed into a very Lara/Player specific class.
isPlayer() is a temporary workaround - my idea about this is to separate this into a Player class based on Character (with Lara/player specific frame(), doAnimCommand() etc. methods - that would make isPlayer() obsolete), but not in this PR... and there's a number of other ideas which I've collected while going through this that may make sense to do before - I'll prepare all my notes and make some proposals once I'm done with this.

stohrendorf:
That's a *very* good reason to keep in mind. It would also allow for multiplayer games once OT is stable enough to extend it with network code :smile:

T4Larson:
I think this is ready to merge now, I've squashed the commits a bit and rebased on the current master, I can merge this if there are no disagreements.
This affects the game/engine loop, entity/character classes (small rearrangements, cleanups) and the animation system.

The good:
- fixed timestep for game logic via bullet's simstep-callbacks (which provides also a catch-up mechanism for (very) slow systems)
   This can be configured for the engine via GAME_LOGIC_REFRESH_INTERVAL, animation stepping can still be
   individually controlled via animation.period (TR_FRAME_RATE by default).
   Currently the Game-Logic/Physics step is 1/60, while the Frame rate is 1/30, ideally GAME_LOGIC_REFRESH_INTERVAL
   should be a multiple of TR_FRAME_RATE.
- a couple of fixes to make animation size and behaviour more consistent with original TR
- rearranged animcommand handling a bit (pre-sorted per Frame/Animm lists, handling centralized in ent:doAnimCommand),
  (Q: I adjusted the Lua functions for this but I couldn't find a non-hackish way to return a Lua table from a Luastate cfunc?)


The bad:
- Collisions are (still) broken, this becomes more apparent in some situations now.
  In some cases you get Lara partially inside a wall, where the penetration fixing fails.
- 2/3click step-climb: Lara may suddenly stand on top of the ledge when doing this:
  This is not a failed animcommand-move, but Lara slightly being moved forward (and set to floorheight) in a game-step(1/60)
  while the climb-up state is accepted and set by the next animation frame(1/30) a gamestep tick later.
  This might also be a collision-detection problem.
- some State transitions must be revised - this applies also the state per-frame callbacks..

I was thinking if I should fix these before proposing this as merge-ready, but the collision stuff may also need some reorganisation
(i.e. ghost-updates without collision updates, which may need either a collision test by bullet, or are redundant and could be remmoved etc.)
and imo should be fixed separately.

The ugly:
- Interpolation glitches (visible in slomo):
  - flip-moves (floor-roll, jump-flip) where an euler-angle turn of 180degrees is ambiguous for quaternion interpolation
  - jump up (from standing) when Lara throws her arms upward
- still a couple of workarounds for lerp adjustment (insta-transforms from i.e. animcommands)

vvs-:
Yes, there are occasional glitches in a jumps, step ups, climbing. Sometimes Lara sinks through the floor. But the importance of this PR should warrant its acceptance. May be somebody will look at the collision bugs? Fixing them before merging would be better.

stohrendorf:
Maybe this is the right time to create a "release" (or "less-buggy") branch, and then merge this PR.

vvs-:
The `master` branch was meant to be that "less-buggy" branch. But people insist to base their experimental work on it and merge it back as soon as possible to lower maintenance costs. On the other hand may be they just subconsciously view `master` as a development branch. In that case creating *release* branch definitely makes sense.

Gh0stblade:
If we split between development/release branches it's almost redundant if you think about it? Everyone has their own forks/branches to develop their own features so it should stay there until it's stable enough to merge into `master`. I'm in favor of keeping `master` as stable as possible.

At this point in time, I don't condone this being merged into `master` just yet, we've seen this happen before when huge changes get merged and when unforeseeable things break, there's a handful that gets upset and starts blaming.

I would only suggest that smaller changes and small bug fixes are merged/pushed into the `master` without thorough reviews/testing until bigger problems are fixed like the one this pull request addresses.

We'll need something like:
1. Testing phase.
2. Bug fixing phase.
3. Final testing phase.
4. Final bug fixing phase.
4. Code review phase.

Until we're 100% certain this won't break too much. As far as I've acknowledged from the notes, this causes collision issues and right now, we cannot afford to have a broken collision system.

Regards.

T4Larson:
> May be somebody will look at the collision bugs? Fixing them before merging would be better.

Yeah, indirectly making collisions worse nags me a bit, so I'm looking into it.
This is clearly a regression for a "release", otoh I'm trying not to get lost too far in my own diverging dev branch, which already changes things that may affect other developmments.
Putting in some quick hacks to get it working seems to me like a step in the wrong direction, but getting it right would imo affect more than just a few fixPenetration calls and have other changes that are not related to the timestep conversion.

Having a "dev" branch would help in cases where one patch may cause regressions in other parts - imo master should be 'clean' and hold the releases, as people usually clone the master branch and expect it to be the best workig version.

> If we split between development/release branches it's almost redundant if you think about it?
Having everyone agree on a bit of procedure can't be an impossible feat, maybe it's time to propose one:
1. Make a PR - don't merge (larger) patches right into dev/master
2. After approval or lack of disagreement, merge to "dev" branch
3. When "dev" has reached a stable condition (milestones, testing), merge into master
4. goto 1

> At this point in time, I don't condone this being merged into master just yet
After thinking about my previous point, I tend to agree - master should have fixed collisions.

As there seem to be no disagreement on the timestep conversion as such, I'll base the collsion dev on it, as a separate branch so depending on how the idea of a dev branch finds acceptance, this can still be merged separately or later as whole.

vvs-:
Good idea. You could even push this PR into a different *development* branch in the official repository. That way anyone else could start fixing the bugs and base their own experiments on it.

vvs-:
This branch is now obsolete. I've tested your T4Larson/develop branch instead.

The monkey swing is fixed. The #266 seems to be fixed as well. Collisions look much better. The #38 seems to be fixed as well. But in that running jump in TR2 she can now reach the ledge apparently but still can't grab it for some reason. Also, there is a new issue when going from swim to wade, Lara now changed to land move type temporary.

T4Larson:
Progress is slow currently, there seem to be new issues creeping up whenever I touch something - i.e. deactivating the sleeping for the kinematic bodies (which is required, or else touching dynamic bodies won't work once they enter 'sleep' state (= green)) not only suddenly blocks the forward wall detection for Lara (she doesn't go into the wall though, but doesn't do the stop-with-hand-at-wall anim either), but also causes lag on level start - which in turn seems to be all the other objects in the level filling their collisionn cache with room-geometry collisions, which again seems to be due to loosely set collision groups, and so on...

The main issue imo is still the awkward ghost-object penetration check, i.e. the speed issues in #266 were due to about 1000 (!) additional full broad/narrowphase checks every tick - just for Laras bodyparts. This appeared to be ok when there are no collisions (because they were discarded early by bullet), but when standing in a simple bounding box like spikes, produces large amounts of (useless) collision points - that's still a problem, even though it seems a bit better now because some redundant collisionchecks are removed; one or two baddies walking around will probably bog this down to 10fps again :worried:

vvs-:
> also causes lag on level start - which in turn seems to be all the other objects in the level filling their collisionn cache with room-geometry collisions

Yeah. I believe that the current fix for that issue introduced in f1db40f5acb8c40e2fdbf12e56f47a85558dcf74 was actually incorrect, i.e. see https://github.com/opentomb/OpenTomb/issues/261#issuecomment-130314165

vvs-:
I'm commenting in this PR for the lack of better place. The T4Larson/OpenTomb@develop has some collision problems, e.g. crumbling floor and boulders fall through geometry. Also, initialization of collisions is slow at the beginning of a level.

T4Larson:
It's a side effect of filtering entity collisions with statics, because most doors etc. are often partly embedded in level geometry - this probably should be handled when activating an entity, but may need some configuration on a per-movable basis: rollingballs etc. should collide with geometry, but doors, switches, spikes etc. should not. 
I'm currently trying to reproduce the effects of the original collision detection with bullet shapes, 

I can't reproduce the slow init, though

vvs-:
> I can't reproduce the slow init, though - are there any specific Levels on which this happens?

Yes, the best one is *The Base* map in TRC. Look for a very slow frame rate at the beginning of the slope.

T4Larson:
There seems to be a lot of entities intersecting each other in this level - I've removed ent/ent collision as well for now, that seems to help with the dynamics init speed...

vvs-:
Now it looks good. Thanks.

vvs-:
Another problem which I noticed recently is that ladder climbing is broken.

T4Larson:
I think the wall climb code had issues even before, because it uses world coords from the sweeptest callback which are not initialized with proper values, and numerous code parts use the same global callback instance for every ray/sweep scan - which derives the resulting world coordinates from it's initial construction values, and hence doesn't work that way when re-used for different rays - this also needs an overhaul.
 I sometimes wonder if a complete rewrite of the entity classes wouldn't be easier...

Issue #181, opened 2015-07-19
Lara can't turn


Lwmte:
Does anybody experience this bug? I'm sure everything worked less than one hour ago, but now Lara refuses to turn (although she leans properly).

vvs-:
At least it works for me in 50275d479f93503dcd2c1d3e499ce25b32333e28.

Lwmte:
Well... Something is wrong on my side then! :smile: 
Although it *looks* like engine wrongly identifies keyboard input as analog input - i. e. boolean value is treated as integer (degree of analog stick turn).

stohrendorf:
Is this still valid?

Lwmte:
No, seems it was some issue with my compiler environment or conflicting commits...

Lwmte:
This is very serious.

I believe, the bug is there because whole controls module was converted to boolean logic without taking move_logic and look_logic operations into account.

vvs-:
I still can't reproduce this.

stohrendorf:
I actually was very careful when working at the controls; but `move_logic` and `look_logic` do the very same as before, as `true` and `false` are converted to 1 and 0 in arithmetic expressions (that's defined in the ISO standard of C++).

Lwmte:
Then I completely don't understand what's happening. It is just very random, but is now much more frequent. Probably really memory corruption issue somewhere?

Lwmte:
Funny thing that when I tried to record a video of it in Fraps, it suddenly vanished one more time.

vvs-:
Perhaps it depends on a frame rate? I can't reproduce bugs which require high frame rate for obvious reason :smile:

Lwmte:
Good idea! Fraps forces engine to work at exactly 60 FPS.

Lwmte:
It seems that I can't reproduce this bug when I build with Release option set in CMake. Perhaps, it's something related to Bullet as well?

Not sure if we should close this or not... I think I will close it for now, until problem resurfaces with release build. There's already too much things to blame debug build for.

Lwmte:
You're not going to believe me, but now I'm getting it with release build.

stltomb:
Is this a 32-bit or a 64-bit build? So far, the only debug/release builds I tested were 64-bit.

stohrendorf:
I'm using 64 builds on Linux and 32 bit builds on Windows (both Debug and Release builds), and cannot confirm this.

Lwmte:
[...]

PR #273, opened 2015-08-14
added vsync screen-configuration option

T4Larson:
Adds a config option to enable/disable vsync.
(If this doesn't do anything for you, please check if you have it forced to on/off in your graphics driver settings)


Lwmte:
Great! It helps to hide my current issue with trashed frame_time values (#181).
Although I was sure that there was no vsync setting in SDL2...

PR #276, opened 2015-08-14
SDL mousemode cleanup, added separate mouse axis scaling

T4Larson:
Removed some unneeded mouse-event handling: SDL_SetRelativeMouseMode already does the pointer hiding/warping and warp-reset filtering by itself, no need to do that in the event handling.

I also added configuration options for the axis scaling (this also allows to easily invert Y-axis via config with a negative scale...) - because it affects the same code part no separate pullreq for that :blush:

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -43,6 +43,8 @@ system =
 controls =
 {
     mouse_sensitivity = 25.0;
+    mouse_scale_x = 0.01;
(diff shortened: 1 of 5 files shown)


Lwmte:
Very neat way to deal with mouse sensitivity, thanks! I think that joystick axis options could be re-made in the same way.

PR #282, opened 2015-08-16
changed camera rotation behaviour: Up/down circles around target

T4Larson:
This changes the camera up/down rotation to a more common behaviour - I'm not sure if I anyone would prefer the current, or if it's just wasting code making this an optional mode?

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -365,6 +365,7 @@ void Game_ApplyControls(std::shared_ptr<Entity> ent)
     cam_angles[1] += 2.2f * engine_frame_time * look_logic[1];
     cam_angles[2] += 2.2f * engine_frame_time * look_logic[2];
 
+    // FIXME: Duplicate code - do we need cam control with no world??
     if(!renderer.world())
     {
         if(control_mapper.use_joy)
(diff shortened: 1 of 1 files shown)


src/game.cpp
```diff
@@ -365,6 +365,7 @@ void Game_ApplyControls(std::shared_ptr<Entity> ent)
     cam_angles[1] += 2.2f * engine_frame_time * look_logic[1];
     cam_angles[2] += 2.2f * engine_frame_time * look_logic[2];
 
+    // FIXME: Duplicate code - do we need cam control with no world??
```
Lwmte: I believe it's a leftover from built-in model viewer, which was removed some time ago (because WadMerger is way easier to use).

I'm not sure it's not completely needed, as theoretically, user can launch engine without pre-loaded level file. However, he won't see anything except splash screen in this case.

vvs-: > theoretically, user can launch engine without pre-loaded level file

Practically, I always do that :smile:

Lwmte:
It's much better this way! We lose ability to "target" entities in debug info mode, as crosshair always points to Lara, but it was tricky anyway. We can always use freelook mode.

vvs-:
> crosshair always points to Lara, but it was tricky anyway

I always disable player to avoid that.

PR #284, opened 2015-08-16
added missing break in fader init

T4Larson:
Found by CDT's code analysis, fixes flashStart() without flashSetup() wrongly doing a fade-to-black.


T4Larson:
Ok.

PR #293, opened 2015-08-24
Animation frame/lerp/time refactoring

stohrendorf:
As I'm not sure whether or not I did something terribly wrong, I'd like to have someone take a look at this to not make things more worse than they currently are in master. From my few tests with TR1, I couldn't see any obvious bugs, though.


T4Larson:
A lot of this interferes with #269, which already fixes some animation oddities (and speeds up loading time as a surprising side effect...but this just in) - there's still a bit of local cleanup to do, I guess I'l have it wrapped up by the weekend.

stohrendorf:
As your work is much more advanced, I'll close this pull request.

PR #298, opened 2015-09-07
[WIP] Split files up into several namespaces

stohrendorf:
This is a WIP PR, meaning that it currently only exists to have a note that this work exists. It currently contains #269 as well. The final goal is to have a sane modularization of OpenTomb, yet there is still a lot of work left. I'm not happy with the current refactoring state, as there are still a lot of things in the wrong namespaces/files (e.g. `world.h`), and many files are still to monolithic (e.g. `animation.h`).


vvs-:
Basing it on #269 was a good idea. Now there should be no complains about merge conflicts.

T4Larson:
Oh...the return of Subdirs and Namespaces.
Just to clarify: My main arguments against this was not the merge hassle.

stohrendorf:
> Moving sources around in loosely chosen nested directories (and namespaces) to pretend some structure which isn't really reflected in the code - this is cargo-cult programming.

This is a WIP. The namespaces already exist, but code still needs to be moved around to get into the right namespaces. And your suggestion of having only two or three namespaces is basically the same as having no namespaces at all, as you soon would be writing `using namespace xyz;` everywhere.

And just to clarify: this is a WIP. Namespaces will be moved/renamed/created/deleted, the same statement holds for files. And if you think about it, it's only natural to put files into folders that are organized just like the namespaces they contain.

It's simply not meant to be merged anytime soon, because it will not be finished anytime soon. Hold your breath. You can still say it's crap when it's finished. Or you can start committing own namespace refactorings just to get me into unresolvable merge hell.

PR #299, opened 2015-09-10
fixed sprite loading for tr4/5

T4Larson:
Top/bottom was flipped, resulting in chunks of gargabe textures in the atlas.
Since Tr4 (and likely Tr5) there is no centering alignment for use as billboard-item anymore (no room sprites), so the l/r/t/b calculation is just a dummy to fill these fields (IIRC, the tx/ty/tw/th match the sprite position/size from the original wad - they're probably a leftover from the level conversion).


vvs-:
I guess somebody should reflect this in TRS3 as well.

Issue #286, opened 2015-08-16
Change default compiling directory


Lwmte:
I reported this bug earlier in VT loader refactoring section, but seems it was ignored.

I believe it is caused by moving large part of CMakeLists.txt from root directory into CMakeLists.txt in /src directory. So we either need to move it back to root script or re-define compiling directory by some CMake command, if it's possible.

vvs-:
RUNTIME_OUTPUT_DIRECTORY

stohrendorf:
...or you simply set the working directory to the source root without adding another hack to the CMakeLists.txt.

Lwmte:
But it worked before, the problem was introduced by separation of CMakeLists.txt. Moreover, I have proper working directory in CodeBlocks properties, most likely it's somewhere in makefile. That's the reason why CMakeLists.txt must be edited.

Lwmte:
By the way, RUNTIME_OUTPUT_DIRECTORY doesn't fix the problem, or I'm using it wrong. I just add this line

`set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "../")`

into CMakeLists.txt placed in /src folder.

vvs-:
No, it's a target property. Should be set with `set_target_properties` command. And there's no `CMAKE_` prefix either.

stohrendorf:
Then it's a CodeBlocks problem, because I'm doing an out-of-source build since I built OpenTomb the first time. And by the way, you should use `"${CMAKE_CURRENT_SOURCE_DIR}/../"` (or was it `CMAKE_CURRENT_SOURCE_DIRECTORY`?).

vvs-:
No, I have the same problem too since the `add_subdirectory`.

stohrendorf:
It works fine for me with QtCreator and MSVC both on Linux and Windows, so it isn't a general problem. Are you *sure* you set the executable's working directory to the source root?

Lwmte:
Here's my CodeBlocks project options:

![clipboard-1](https://cloud.githubusercontent.com/assets/3331699/9295733/0d38c2aa-4481-11e5-9893-2bb3985fa6ad.png)

And here's where binary is placed:

![clipboard-2](https://cloud.githubusercontent.com/assets/3331699/9295734/0d3dac3e-4481-11e5-9cca-c3be2237ed6a.png)

vvs-:
The executable is always created in the same directory level as `CMakeLists.txt` file. It's a documented behavior.

[...]

PR #300, opened 2015-09-10
cmake: added build type selection, build exe in base dir

T4Larson:
Setting the build type will generate the appropriate optimize-flags for compilation.
I also changed the build directory for the executable (was in src/, is now in the base dir, where it can be directly run)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,11 @@ cmake_minimum_required(VERSION 2.8)
 
 project(OpenTomb)
 
+if(NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE Debug CACHE STRING "Build type" FORCE)
+    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS Debug Release MinSizeRel RelWithDebInfo)
+endif()
+
 if(APPLE)
     file(GLOB OPENAL_SRC
         3rdparty/al/*.c
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -88,6 +88,8 @@ target_link_libraries(
     # efence
 )
 
+set_target_properties(${PROJECT_NAME} PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR})
+
 if(CMAKE_COMPILER_IS_GNUCC)
     target_compile_options(${PROJECT_NAME} PRIVATE -std=gnu++11 -Wall -Wextra -msse3)
 elseif(MSVC)


T4Larson:
```
if(NOT CMAKE_BUILD_TYPE)
```
> This is not necessary, as it's already handled by cmake automatically.

Not sure what you mean - the condition which you marked is needed, otherwise cmake will reset the build type to the default (debug) every time you configure/generate.

> This is a hack which is already discussed in #286.
Ah, I see, but not much of a solution there.
As this target property seems the officially documented way for this purpose and I'm using documented and well defined variables for the output directory, it can hardly be labeled as a hack.

stohrendorf:
> the condition which you marked is needed, otherwise cmake will reset the build type to the default (debug) every time you configure/generate.

Then it's probably an IDE issue, setting the build type explicitly, as the build type is already cached by cmake itself.

About the output directory: Yes, you *can* set it this way if you have very specific requirements, but it's a bad habit to do so. It's even discouraged to do in-source builds (using `cmake .`), and this hack won't work if you do the preferred out-of-source builds, so it's a non-working hack, too. It *can* be used when the project contains shared objects (aka DLLs/SOs) which are dynamically loaded from a path relative to the executable, so you can build and run it without installing it first to a local directory, but then you get into trouble setting up RPATHs and such. I did it once [here](https://github.com/stohrendorf/ppplay/blob/master/CMakeLists.txt#L87), and it's still a mess. It basically narrows the freedom of chosing the build structure to a very specific layout. It can be seen as a convenient build/test setup (which I did back then), but usually it creates more trouble than it solves (e.g. messing up the repository by accidentally committing build files, or putting too many regex's into `.gitignore`, leading to missing files/changes, like it already happened).

A *clean* way to handle this is to make the root path configurable, e.g. by only expecting `config.lua` (or `autoexec.lua`) in the current directory, and set up the scripts' root path there.

T4Larson:
> Then it's probably an IDE issue, setting the build type explicitly, as the build type is already cached by cmake itself.

Specific IDE issues are irrelevant - Cmake may ignore this when you generate configurations for environments or IDEs which can handle multiple build types.
But remember that Cmake can generate a large variety of configurations, which do not necessarily offer the same flexibility as your preferred IDE: Makefiles are probably the most simple case, Cmake can't magically guess-and-cache the build type for these.

> ... and this hack won't work if you do the preferred out-of-source builds, so it's a non-working hack, too.

That's wrong.
Assuming you have configured cmake with "mybuild/" as build directory, then the executable will be built in "mybuild/opentomb.exe" (instead of in "mybuild/src/opentomb.exe") - I cannnot see how that breaks an out-of-source build.
Maybe you could just try this first, so we can keep the discussion to non-speculative issues.

stohrendorf:
> Specific IDE issues are irrelevant [snip]

They're irrelevant in the sense that OpenTomb does not need to care about it, that's right.
> Makefiles are probably the most simple case, Cmake can't magically guess-and-cache the build type for these.

Wrong. When you call `make`, cmake calls itself to check if the CMakeFiles have changed, and then re-configures itself if necessary. If you change your build type, the makefiles are changed, too, to integrate other compiler flags.

Prove:

```
sto@linux-shba ~/devel/OpenTomb-build $ make
[100%] Built target OpenTomb
sto@linux-shba ~/devel/OpenTomb-build $ touch ../OpenTomb/CMakeLists.txt
sto@linux-shba ~/devel/OpenTomb-build $ make
-- Found libsndfile: /usr/lib64/libsndfile.so
-- Configuring done
-- Generating done
-- Build files have been written to: /home/sto/devel/OpenTomb-build
[100%] Built target OpenTomb
sto@linux-shba ~/devel/OpenTomb-build $ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ../OpenTomb
-- Found libsndfile: /usr/lib64/libsndfile.so
-- Configuring done
-- Generating done
-- Build files have been written to: /home/sto/devel/OpenTomb-build
sto@linux-shba ~/devel/OpenTomb-build $ make
[  1%] Building CXX object src/CMakeFiles/OpenTomb.dir/common.cpp.o
[  3%] Building CXX object src/CMakeFiles/OpenTomb.dir/gl_font_buffer.cpp.o
[  5%] Building CXX object src/CMakeFiles/OpenTomb.dir/main_SDL.cpp.o
```

> I cannnot see how that breaks an out-of-source build.

I thought you were trying to "fix" in-source builds, sorry.

T4Larson:
>  Wrong. When you call make, cmake calls itself to check if the CMakeFiles have changed, and then re-configures itself if necessary.

What I said is correct, and also answers your question why setting a build type is useful.
We don't need to discuss obvious details about cmake's configuration checking.

> I thought you were trying to "fix" in-source builds, sorry.

See the commit msg.
The argument about hackishly ruining custom build directories doesn't seem to apply, and since this patch apparently also resolves issue #286, I see no problems with it.

stohrendorf:
Well, "base dir" is ambiguous: source (base) dir or build (base) dir; I had the first in mind, as it was a pretty memorable discussion.

But I still don't understand the configuration system. I don't know of *any* build system cmake supports that "automagically" changes the build configuration type, and I never heard of such cases (but I've seen config files using this "caching trick", resulting in unexpected build type changes). With which build system does this happen?

T4Larson:
"magically guess-and-cache" seemed to be the implication of "not necessary to set explicitely because it is cached" when you create a build configuration where this is never set to be cached i the first place (other than the default).

> But I still don't understand the configuration ~~system~~ problem.

If you generate project files for an IDE which manages multiple build types, then setting the build type in cmake won't make a difference, because you're selecting the build type in your IDE.

But for i.e. makefiles you would need to specify the desired build type somehow.
You could either
  - need to have build-type specific targets generated in the makefile, for each build-type variant - Cmake doesn't seem to do this.
  - crudely override compiler specific options for make - you don't want to do this.
  - you do it like done here: You specify it explicitely for Cmake (either on the cmake cmdline, or as convenient selection in cmakelists), so the makefile is generated for the chosen build type.

I'd think that you will probably appreciate the side effect that this way, you can have different build directories for each build type, even with classic makefiles.

stohrendorf:
You can set a default build type using:
```cmake
if (NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE "Release")
endif()
```

On the other hand, your third point *is* handled by cmake, i.e. with `cmake -DCMAKE_BUILD_TYPE=Debug ...` the build type is cached to be `Debug`, and can be changed to other build types using the same command line. And if explicitly set, it's cached:

cmake-build-type

T4Larson:
>You can set a default build type using:
if (NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE "Release")
endif()
...

I'm glad to explain why and what I did, but this is ridiculuous - are you trying to waste my time?

stohrendorf:
No, I'm not. I'm seriously trying to understand what you want to achieve and why. The thing is that cmake already caches the build type if set, and if not, the last three lines of code do the trick. You don't need to mess around with the internal build type caching mecahnism.

If I understand you correctly, all you want to achieve is a default build type. Put the three lines I posted somewhere at the top of the main CMakeLists.txt, and you're done. There won't be any automatic changes in build type configuration, as I used this trick myself a few times.

T4Larson:
- You're suggesting an inferior version of what I did in this patch (no type selection, no default setting). I have the impression you haven't even looked at the patch, not to mention actually testing it.
 - Your's doesn't work; Create a _clean_ build configuration for a makefile project with it (or delete the cached build_type entries) and check the cached default build type value: It's an empty string (cmake 3.3.0, I do not assume platform specific differences in this behaviour)
 - You're implying hackish cruft again ("mess around with the internal build type caching mecahnism"). This is not the case, RTFM.
 - I have a cool dropdown in the gui, to conveniently select from available build types, you don't.

stohrendorf:
* I have taken a look at the patch. And because your commit message and PR description are ambiguous and indicate "hacked" behaviour, this discussion started. The base dir is ambiguous in terms of source base dir and build base dir. And the build type selection/caching is already provided by cmake itself. But I have to admit that I missed setting the default build type because of the noise, which is the only relevant change in the commit (see next point).
* cmake uses the overridden build type when generating, but doesn't store it when not explicitly set; isn't that the meaning of having a default value, setting one if none is provided?
* It's redundant because of the point I mentioned before this. You're replicating already implemented behaviour, and then change its behaviour slightly. I call this hacking.
* Is that really needed? OK, it's *sometimes* convenient, but how often do you change build types?

T4Larson:
You keep repeating already refuted arguments. I see no constructive value in discussing this any further with you.
The "need" to have the default shown and easily selectable in the gui may be arguable to a degree; if you can't remember "RelWithDebInfo" this certainly helps and avoids typos - it's a comfort feature.

CMakeLists.txt
```diff
@@ -7,6 +7,11 @@ cmake_minimum_required(VERSION 2.8)
...
 project(OpenTomb)
 
+if(NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE Debug CACHE STRING "Build type" FORCE)
+    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS Debug Release MinSizeRel RelWithDebInfo)
+endif()
+
 if(APPLE)
     file(GLOB OPENAL_SRC
         3rdparty/al/*.c
```
TeslaRus: need to be updated, folder was renamed

stohrendorf: ... and it's the `develop` branch, which still contains the old refactored code base. In the new "phoenix" code base, there is no build type needed at all, as the `CMakeLists.txt` is only used for asciidoc invocation, not for building executables.

vvs-: > there is no build type needed at all, as the CMakeLists.txt is only used for asciidoc invocation, not for building executables

That's just temporary, while there is no cross-platform build system.

TeslaRus: that request for other file, new one is differ:
```
# Usage (Linux):
#   $ cd OpenTombSrc
#   $ mkdir build && cd build
#   $ cmake .. && make

cmake_minimum_required(VERSION 2.8)

project(OpenTomb)

message(WARNING "This CMakeLists.txt does not generate executables, it is only used to process the documentation.")

add_subdirectory(doc)
```

vvs-: But I should repeat again: it's a temporary measure, just to be able to build the documentation. In a few days we should find a solution for a working building system. The previous one was broken and created too much controversy. That doesn't mean that this PR is still valid, though.

PR #318, opened 2016-06-22
Build fixes

vvs-:


src/core/system.c
```diff
@@ -1,6 +1,7 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <time.h>
```
TeslaRus: is it really needed? there is #include <sys/time.h>
strange... interesting what is GCC version?

vvs-: Yes, it doesn't build on Linux without these fixes. This was tested with gcc 5.3.1 and this depends not on the compiler but on glibc which is different on Linux. In fact, it could be even insufficient on any other OS. I can't test it on Windows, so I don't know if this breaks anything there.

Gh0stblade: Windows build is working with no issues without that header. Has anyone been able to check if Mac works? @cochrane

vvs-: Yes, but does it work _with_ it? It worries me that while it works flawlessly on Linux, it could be broken on any other OS. And I have no means to test it.

Gh0stblade: Yes, building under windows fine.

stohrendorf: I just requested access for Travis CI in OpenTomb, so that these questions shouldn't arise anymore once a suitable build script exists. So I guess a proper cross-platform build system which allows to compile OpenTomb (but doesn't need to create deliverable executables in the first shot, as it's only for automated compile tests) has some sort of priority.

PR #319, opened 2016-06-22
[NOT TO BE MERGED] Just some example project

vvs-:
This is NOT intended for merging. It's just a quick hack and a starting point for discussion. Also it could be useful for those adventurers who want to quickly try the new code on Linux.

The proposed build system should be more complete than the old system was. The best starting point should be @stohrendorf build system based on Hunter.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,12 +1,176 @@
 # Usage (Linux):
-#   $ cd OpenTombSrc
-#   $ mkdir build && cd build
-#   $ cmake .. && make
+#  MKDIR build folder next to opentomb folder (e.g., opentomb-build)
+#  CD into build folder
+#  cmake ../opentomb
+#  make
 
-cmake_minimum_required(VERSION 2.8)
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
 
-project(OpenTomb)
+PROJECT(OpenTomb)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,12 +1,176 @@
...
+#  make
 
-cmake_minimum_required(VERSION 2.8)
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
 
-project(OpenTomb)
+PROJECT(OpenTomb)
 
-message(WARNING "This CMakeLists.txt does not generate executables, it is only used to process the documentation.")
+SET(CMAKE_CXX_FLAGS "-std=c++0x")
```
stohrendorf: Maybe bump that up to `c++11` or even `c++14`.

stohrendorf:
The downside of hunter is that it requires an external file to add new packages that aren't already provided by hunter. Also, I have a few hacked (forked) repos of popular libs to make them compatible with hunter. That's maybe not such a good idea.

vvs-:
The OpenTomb established workflow depends on local in tree libraries. So, it's not that bad in comparison. Also, you already used a macro that can select between system and Hunter libraries. It should work well in practice. We should test it on OS X and Windows to assess the consequences.

stohrendorf:
From what I can tell it both works on Windows and Linux, but I can't say anything about OS X. Especially boost (if used) takes hours (!) to build on Windows, and I usually prefer to download a pre-built boost dev package and force the project to use that.

PR #320, opened 2016-06-23
Travis CI script and very basic CMakeLists.txt

stohrendorf:
Provides a basic Travis CI script, yet it doesn't build because of platform-specific (Linux) problems. See https://travis-ci.org/stohrendorf/EdisonEngine/builds/139702888 for a few of the problems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: cpp
+compiler:
+  - clang
+  - gcc
+sudo: required
+dist: trusty
+before_install:
+  - sudo apt-get -qq update
+  - sudo apt-get install -y cmake libopenal-dev libsdl2-dev libsdl2-image-dev liblua5.2-dev zlib1g-dev
+install: true
+script:
+  - mkdir build
+  - cd build
+  - cmake ..
+  - make
+
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,366 @@ cmake_minimum_required(VERSION 2.8)
 
 project(OpenTomb)
 
-message(WARNING "This CMakeLists.txt does not generate executables, it is only used to process the documentation.")
+message(WARNING "This CMakeLists.txt does not generate usable executables, it is only used to process the documentation and test compilability.")
 
 add_subdirectory(doc)
+
+if(CMAKE_COMPILER_IS_GNUCXX)
+  message(STATUS "GNU C++-like compiler detected")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -Weffc++")
+endif()
+
+if(CMAKE_COMPILER_IS_GNUCC)
+  message(STATUS "GNU C-like compiler detected")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99 -Wall -Wextra")
+endif()
+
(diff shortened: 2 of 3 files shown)


vvs-:
What I see is that there are too many warnings caused by too strict C++ switches in compiler. The log file exceeded quota and compilation failed. There is `std=c++11` in CMakeLists.txt, apparently this is too much for `master`.

vvs-:
It compiles successfully on my computer using this CMakeLists.txt with many warnings, mostly from Bullet. Also, it does _not_ use in tree OpenAL and uses system library instead. The linking failed due to some missing libraries (I didn't investigate which ones in particular, but looks like SDL_image was involved).

stohrendorf:
It really needs some love. And I guess it was merged too soon again :/

vvs-:
Yes, but nothing bad could happen this time because there was nothing to broke ;) There was no working CMakeLists.txt anyway.

stohrendorf:
OK, I agree. But it feels dirty :)

vvs-:
Don't count on opinions because it's subjective. Count on experiment because it's scientific. We should make decisions based on testing and testing only :eyeglasses:

TeslaRus:
In very old build script version I build "extern" libs to static libs and in final stage link them (GCC linker fails if object list is too big);

stohrendorf:
Some admin still needs to enable the automated build from within travis-ci.org.

TeslaRus:
Instruction please!

stohrendorf:
Go to travis-ci.org, authenticate with your github account, and enable it for the OpenTomb repo. On travis-ci.org, there is some sort of step-by-step guide that should help you.

TeslaRus:
ok, do I need to push .travis.yml again?

stohrendorf:
No, it's merged. Travis will now be triggered on every push and use the current .travis.yml to build. See here: https://github.com/opentomb/OpenTomb/pull/321

vvs-:
There are several problems with this CMakeLists.txt:
- There are compiler flags `-Wall -Wextra -Weffc++` which are too noisy when compiling Bullet. This exceeds Travis 4 MB quota for its logs.
- OpenTomb requires OpenGL libraries for linking.
- It can't link in-tree libvorbis. This library rely on autotools to be properly configured. The included files were pregenerated in some particular setup which was different from Linux build environment. It actually links with my CMakeLists.txt by pure chance. That might lead to some hard to discover memory allocation bugs at run-time.

TeslaRus:
Engine uses OpenGL but do not requires linking OpenGL, It uses SDL_GL_GetProcAdress; In ENGINE.md described: do not use any gl* function, use qgl* functions, defined in core/gl_util.h
"It can't link in-tree libvorbis." - if I have more time and crazy perseverance, I uses SDL function and own written platform independent realizations of all;
+I do not like to add new libs to project, code become like ZOO + arise troubles like this;
question: cygwin provides programming environment like Linux / OSX? The code written in cygwin environment will works equally in different platforms?

vvs-:
> Engine uses OpenGL but do not requires linking OpenGL
Yes, I was confused by its presence in the previous cmake project.

> question: cygwin provides programming environment like Linux / OSX? The code written in cygwin environment will works equally in different platforms?
I doubt it. It has good POSIX compliance though, much better that Windows own libraries.

TeslaRus:
ok, I will write to write in forum about cygwin usage under MSW and using POSIX for better crossplatform compatibility;

PR #324, opened 2016-06-23
Make Travis happy

vvs-:
There were three actual reasons:
- compiler warnings
- `std=c99` flag prevented linking of libvorbis
- no SDL2_image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,18 +13,20 @@ add_subdirectory(doc)
 
 if(CMAKE_COMPILER_IS_GNUCXX)
   message(STATUS "GNU C++-like compiler detected")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -Weffc++")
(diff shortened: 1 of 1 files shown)


stohrendorf:
Well, not entirely, unfortunately. OpenTomb itself has to be compiled with `--std=c99`, while the external libs mustn't. So either the external libs must be wrapped into `add_library`, or the system provided libs must be used.

vvs-:
It doesn't require it. I've built OpenTomb executable and it's running successfully.

stohrendorf:
But Travis isn't happy :)

vvs-:
How do you know? It still using old project.

BTW, there are no C files in OpenTomb, so how it could be affected?

vvs-:
Who wants to fix that? There are several ways. It can be fixed on Travis side, on OpenTomb side, or in this CMakeLists.txt. I won't be able to work on it until tomorrow, but this branch is open for any developer to add commits to it.

TeslaRus:
OK, if You want to fix autobuild, You can push it without my resolution (CMakeLists.txt, .travis.yml and other autobuild script), but I suggest something like vvs make fix, stohrendorf check it, and stohrendorf make fix, vvs check it...

PR #326, opened 2016-06-24
Separate core

vvs-:
Should work now

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,16 +18,6 @@ set(OPENTOMB_SRCS
     src/audio.cpp
     src/character_controller.cpp
     src/controls.cpp
-    src/core/console.c
-    src/core/gl_font.c
-    src/core/gl_text.c
-    src/core/gl_util.c
-    src/core/obb.c
-    src/core/polygon.c
-    src/core/redblack.c
-    src/core/system.c
-    src/core/utf8_32.c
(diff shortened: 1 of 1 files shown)


vvs-:
The old gcc in Travis doesn't compile OpenAL. We need to use system library, there is no choice.

vvs-:
Pull requests doesn't work that way. They should be amended until they work and this one doesn't.

CMakeLists.txt
```diff
@@ -65,12 +55,12 @@ set(OPENTOMB_SRCS
...
 add_definitions(-DOV_EXCLUDE_STATIC_CALLBACKS)
 
-add_executable(OpenTomb ${OPENTOMB_SRCS})
+add_executable(OpenTomb ${OPENTOMB_SRCS} $<TARGET_OBJECTS:core>)
 
 if(CMAKE_COMPILER_IS_GNUCXX AND CMAKE_COMPILER_IS_GNUCC)
     message(STATUS "GNU-like compilers detected")
     # Trimming the string may result in a compilation error
-    set_property( TARGET OpenTomb APPEND_STRING PROPERTY COMPILE_FLAGS " -std=c++11 -std=c99 -Wall -Wextra ")
+    target_compile_options(OpenTomb PRIVATE "-std=c++11")
```
stohrendorf: Thanks for cleaning it up :)

PR #327, opened 2016-06-24
[NOT_FOR_MERGING] Fix OpenAL

vvs-:
This commit should be in the previous PR, not a separate one.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ if(CMAKE_COMPILER_IS_GNUCXX AND CMAKE_COMPILER_IS_GNUCC)
     # Trimming the string may result in a compilation error
 endif()
 
-target_compile_options(OpenTomb PRIVATE -std=c++11 -std=c99)
+target_compile_options(OpenTomb PRIVATE $<$<COMPILE_LANGUAGE:CXX>:-std=c++11> $<$<COMPILE_LANGUAGE:C>:-std=c99>)
 
 include(FindPkgConfig)
 pkg_search_module(SDL2_IMAGE REQUIRED SDL2_image)
@@ -378,7 +378,7 @@ set(CORE_SRCS
 
 add_library(core OBJECT ${CORE_SRCS})
 target_include_directories(core PUBLIC extern/freetype2/include ${LUA_INCLUDE_DIR})
-target_compile_options(core PUBLIC "-std=c99")
+target_compile_options(core PUBLIC $<$<COMPILE_LANGUAGE:C>:-std=c99>)
 
 add_library(bullet STATIC ${BULLET_SRCS})
 target_include_directories(bullet PUBLIC extern/bullet)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,8 +60,9 @@ add_executable(OpenTomb ${OPENTOMB_SRCS} $<TARGET_OBJECTS:core>)
 if(CMAKE_COMPILER_IS_GNUCXX AND CMAKE_COMPILER_IS_GNUCC)
     message(STATUS "GNU-like compilers detected")
     # Trimming the string may result in a compilation error
-    target_compile_options(OpenTomb PRIVATE -std=c++11 -std=c99)
 endif()
+
+target_compile_options(OpenTomb PRIVATE -std=c++11 -std=c99)
 
 include(FindPkgConfig)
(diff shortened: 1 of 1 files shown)


vvs-:
Please, do not merge any PR until it is considered working.

I just don't understand why there are always different results on Travis? I have no such errors with Lua. How am I supposed to test it then?

vvs-:
The differences in Travis build system are just too big. I didn't get these errors even on my old system where it produced similar errors in `core`. Wouldn't it be easier to just upgrade to a more recent build environment on Travis? @stohrendorf, was there some reason why you chose that particular Ubuntu version?

vvs-:
Great, gcc build works now, but clang doesn't. I have no means to fix that. I propose to turn it off in travis configuration. Any objections?

stohrendorf:
@vvs I remember a discussion about system requirements, and the result of it was something like "we need to make sure it runs on old and weak systems", so Ubuntu 14.04 already seemed like a risky choice ;)

stltomb:
@vvs- It doesn't compile because the check `if(CMAKE_COMPILER_IS_GNUCXX AND CMAKE_COMPILER_IS_GNUCC)` only works for gcc, and not Clang.
Because of that, the compiler options like _-std=c++11_ aren't supplied to Clang.

[see here](https://stackoverflow.com/questions/10046114/in-cmake-how-can-i-test-if-the-compiler-is-clang)

vvs-:
> "we need to make sure it runs on old and weak systems"

"Runs" and "builds" are different things. Shouldn't the whole purpose of Travis CI be in compilation alone?

> It doesn't compile because the check if(CMAKE_COMPILER_IS_GNUCXX AND CMAKE_COMPILER_IS_GNUCC) only works for gcc, and not Clang.

Good catch. But I still won't be able to test it locally :frowning:

stohrendorf:
> "Runs" and "builds" are different things.

Not necessarily. If you require a very recent compiler (or even libs) to build OpenTomb, chances are pretty high that it won't run on older systems because of system library dependencies. So if you want to ensure that it (most probably) will run on older systems, you need an older system to compile it.

vvs-:
That's impossible to support. The question was: is it reasonable to require users to update their systems in order to be able to run OpenTomb? And my position was that it should be a requirement _only_ if it's _absolutely_ necessary for an OpenTomb to run. There are requirements that can't be justified on these grounds.

vvs-:
> I'd rather fix the if than provisioning these flags generally (and maybe also remove that obsolete comment). If someone (like me) wants to try out MSVC some day, this will prevent a compilation.

And you are free to do that. But I'm unable to test such changes, so don't expect me to do that blindly.

vvs-:
`Error evaluating generator expression:

    $<COMPILE_LANGUAGE:CXX>

  Expression did not evaluate to a known generator expression`
So, it's impossible to fix without upgrading Travis.

stohrendorf:
According to https://cmake.org/cmake/help/v3.0/variable/CMAKE_LANG_COMPILER_ID.html that expression should be something like
```cmake
if((CMAKE_COMPILER_IS_GNUCXX AND CMAKE_COMPILER_IS_GNUCC) OR (CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
```

vvs-:
That's right, but I can't test it. If there will be a typo I can't notice that and it would silently fail on a user system.

stltomb:
> So, it's impossible to fix without upgrading Travis.

@vvs- It would seem so (without using the `CMAKE_CXX_FLAGS` and `CMAKE_C_FLAGS`).

That CMake version must be ancient. Trusty repos have version 2.8.12 that was released on Oct 8, 2013. The `COMPILE_LANGUAGE` feature was merged on March 9, 2015.

vvs-:
Could you change Trusty to something more recent? I believe that it's completely justified in this case.

vvs-:
Looks like the only way to solve this is to install CMake 3.3+ from PPA.

vvs-:
Success! :fireworks:

But this is a nasty workaround and we should find a better solution.

vvs-:
Feel free to merge this PR if there are no other objections.

stohrendorf:
It's a first working build system that make Travis happy, so if this gets merged, Travis can actually tell if the pull requests are sane or not, which is a huge step forward. I'd go for it, but keep in mind that it needs cleanup.

Issue #275, opened 2015-08-14
Oneshot Trigger behaviour


T4Larson:
Tr1: Switch oneshot should lock the item only if the object is triggered on, in the original engine a oneshot-switch doesn't seem to lock an object if the os-switch only switches it off.

tr4: Tr4 has 3 separate oneshot-lock-bits for trigger,antitrigger and swtch types - that means activating a oneshot-trigger will still allow you to antitrigger and switch the objects.
There's one special case (due to a fallthrough in the code), where a os-switch set both the switch-lock-bit _and_ the trigger-lock-bit, when that oneshot-switch activates an object (if the os-switch deactivates an object, only the switch-lock-bit is set).
Tr4 has also a "activation-state-negated" bit, which is used for pre-triggered objects: When pre-triggered objects are loaded (mask=11111), then the negation flag is set and the mask is cleared - so it must be triggered with all bits again to be **de**activated.
This bit is also cleared by the Tr4 antitrigger: Pre-triggered objects will be deactivated, and not be negated anymore from this point on.

Lwmte:
1. I got understanding of this setup through HUGE frustration with TR1 Gold one-shot secret triggers (thanks to amount of testing done by @vvs-). I understood it this way: if entity was *not* yet activated, then one-shot flag makes no difference (no idea if it's set by trigger which not yet activated it), and if entity is being activated, then one-shot flag takes action. In OpenTomb it's currently done via checking current entity activation mask - if it's fully set (11111), and one-shot flag is set as well, trigger processing is bypassed. You can learn the way it works in `trigger_functions.lua` at lines 64-68.

You can also see that there's a condition for `TR_III` game version - the reason is, during my tests with TR3 I realized that *any* one-shot trigger locks entity - even if it doesn't fully set activation mask. For example, I have two triggers for some entity, one with mask 00001 and one-shot flag set, and another with mask 11110 and one-shot mask not set. If I step on first trigger, and then on second, entity won't activate! But if I step on second trigger, and then on first, entity will activate.

2. So it's actually works in such complicated way? I thought that *antitrigger* just resets one-shot flag along with activation mask for a given entity. But now I did additional tests, and it's much more clear to me. However, this setup requires major changes in trigger layout for each entity... Current approach must be changed completely (as I used packed bitfield for keeping activation mask, oneshot flag and activation state flag in the same 8-bit field).

[...]

PR #331, opened 2016-06-24
fixed sharp edges climbability checking;

TeslaRus:
found that "middle hands pos" may contains too low z coordinate in some
animations; the solution is not full: 
- not all cases tested; 
- Lara can climb on through a wall;
- Character_GetMiddleHandsPos have to return point with z_max, not with z_middle;

### DIFF
--- a/src/anim_state_control.cpp
+++ b/src/anim_state_control.cpp
@@ -1853,6 +1853,7 @@ int State_Control_Lara(struct entity_s *ent, struct ss_animation_s *ss_anim)
                 climb_to[0] = climb_from[0] + t * ent->transform[4 + 0];
                 climb_to[1] = climb_from[1] + t * ent->transform[4 + 1];
                 climb_to[2] = climb_from[2] - ent->character->max_step_up_height;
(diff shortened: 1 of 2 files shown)


vvs-:
I can't find a test case for this issue. Could you provide one, please?

TeslaRus:
to check Character_CheckClimbability in unit test: test level and some cases of test_from, test_to are required; returned collision_info will be criteria; the level have to contain some climbable edges and character entity (really needed only character settings...);

vvs-:
So, are there no such existing levels in TR? If not then perhaps someone could create a test level? @Lwmte, could you make one, please?

Also, there were some test levels in `backup` which need to be resurrected.

TeslaRus:
ok, this time I trying to build OT with cygwin, and SDL2 makes a problem for linker (sdl redefines main, so linker troubles); ok, I will see test levels from backup, and one of them will be used as a default (empty engine looks not so good);

vvs-:
There is another test level which is buried in issues: https://github.com/opentomb/OpenTomb/issues/275#issuecomment-131443375

PR #336, opened 2016-06-25
Remove Mac OS-specific #ifdefs

cochrane:
These code pieces tried to do two different things, neither of which makes any sense:

-   Set up the current working directory to some external location to find files there
    more easily. This uses a function that is not actually in the repository, and can thus
    never work.

-   Avoid using SDL_image, because there were special code paths for OS X to use CoreImage
    for saving. This code path is gone (and I think they should stay gone), so this lack of
    initialization is definitely wrong.

This means mostly that the OS X version now uses the same source codes as all other
versions, which makes an awful lot of sense. However, it does not mean the Mac version works
yet.

There are no effects on other platforms.

### DIFF
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -99,10 +99,6 @@ void ShowDebugInfo();
 
 void Engine_Start(const char *config_name)
 {
-#if defined(__MACOSX__)
-    FindConfigFile();
-#endif
-
     Engine_InitDefaultGlobals();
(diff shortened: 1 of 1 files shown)


TeslaRus:
I found using (configure and compile) SDL_Image not convenient, so first step - I want to create in engine little image module with abstract interface, but in implementation may be self - written image loaders, sdl image, direct usage libpng, libjpeg, or other libs usage;

PR #348, opened 2016-06-27
Switch to system OpenAL throughout.

cochrane:
This completely removes the custom OpenAL implementation and all its weird side-effects, and instead uses the
system-provided one. The system-provided one has a few differences, and these changes are all about mitigating
the effects. That means:

1.  The MacOS version of OpenAL uses headers of the form `<OpenAL/al.h>` instead of the more common `<AL/al.h>`. This code switches to `<al.h>`. CMake ensures that the paths will still work.

2.  The MacOS version of OpenAL does not come with `alext.h`, `efx.h` and `efx-presets.h`. To work around this, `CMakeLists.txt` now contains a check for these files and automatically generates a config file (based on the template in `config-opentomb.h.in` with the results. The code in `audio.cpp` that makes use of these features is then hidden behind a lot of `#ifdef`s.

## Result

The code compiles on Mac OS X. It may also run; I haven't tested it yet.

## Drawbacks

-   The special sound effects won't work on Mac OS X. This is a fundamental limitation of the system-provided OpenAL library that can only be worked around by including our own (customized or not). I don't see it as a showstopper at the moment; I mostly play with sound off anyway.
-   The include file path changes may cause problems for users of IDEs who have generated their project files manually, instead of using CMake to do so.
-   `audio.cpp` now depends on the existence of a file that is generated by CMake on-the-fly based on system configuration. **That means it's impossible to use an IDE with a manually generated project file**, because that file will be missing (there are possible hacked workarounds, by commenting out the include and defining the macros manually).

This is overall a fairly major change, and I would advise against merging it until everyone understands and agrees with the repercussions. If the consensus is to ship our own OpenAL (customized or not), then this entire pull request becomes void anyway.


vvs-:
Works for me. And Travis too :smile:

cochrane:
With b4103c2, there is a default version of the configuration file for those who don't want to use CMake. So I think this should now work for everyone (modulo maybe some setup work).

TeslaRus:
I will try to build properly (make working) build-in OpenAL in my Linux machine without win build changing. I may merge this request but without internal/AL deleting.

stohrendorf:
@TeslaRus Please don't keep the sources when merging. I see the advantages of keeping them, but the team is too small to keep track of library updates and custom build solutions for it. And it also eliminates the risk of building OpenAL the wrong way and then hunting gremlins.

PR #349, opened 2016-06-28
Add the headers of our files to CMakeLists

cochrane:
The header files do not need to be compiled, so it may not seem like they need to be in CMakeLists.txt, but that is wrong. Changes to them still trigger recompiles elsewhere, so CMake needs to be aware of them to include them in the generated dependency tracking. Also, without this, the headers aren't in the generated IDE files, which makes life surprisingly annoying.

This is completely independent of #348, and I expect no issues with merging this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,41 +15,72 @@ add_subdirectory(doc)
 
 set(OPENTOMB_SRCS
     src/anim_state_control.cpp
+    src/anim_state_control.h
     src/audio.cpp
+    src/audio.h
     src/character_controller.cpp
+    src/character_controller.h
     src/controls.cpp
+    src/controls.h
     src/engine.cpp
+    src/engine.h
+    src/engine_string.h
     src/entity.cpp
+    src/entity.h
     src/game.cpp
+    src/game.h
     src/gameflow.cpp
+    src/gameflow.h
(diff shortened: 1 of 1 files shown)


stohrendorf:
Maybe you'd like to also include [this function](https://github.com/stohrendorf/EdisonEngine/blob/master/CMakeLists.txt#L28) to have a nice folder-based grouping in VS-based project files. Just call `group_files(${SOURCES})`, and you're set.

vvs-:
I've seen that CMake automatically generate dependencies. I never checked their effect on compilation though. So, I don't argue that headers _might_ need to be included.

cochrane:
@vvs- To be honest, I'm not sure they're actually needed for dependency tracking, I just heard a rumor like that on the interwebs. But I am certain that they're needed so they'll show up in IDEs (e.g. Xcode, but probably all the others as well).

@stohrendorf I don't think that works with Xcode, so I probably won't add it, but if someone else wants to have a go, this does seem like a good idea.

stohrendorf:
Although build systems *should* track dependencies, some (few) project file based systems don't do that properly. But it's really useful to see them in the project tree.

@cochrane It's a VS-only grouping thing, but it's extremely helpful there; as VS is probably not a target IDE, it was just a suggestion.

PR #351, opened 2016-06-28
Add proper newlines to debug log.

cochrane:
The previous code attempted to add a newline at the start, which then got immediately overwritten by the
following code. The new code adds a newline at the end, which actually works. This was just a minor thing that annoyed me.

(cherry picked from commit c95f7fd49efb500f849815be24eccf37e08e89b2)

### DIFF
--- a/src/core/system.c
+++ b/src/core/system.c
@@ -170,11 +170,21 @@ void Sys_DebugLog(const char *file, const char *fmt, ...)
     va_list argptr;
     static char data[4096];
     FILE *fp;
+    size_t len;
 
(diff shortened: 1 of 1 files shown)


src/core/system.c
```diff
@@ -170,22 +170,32 @@ void Sys_DebugLog(const char *file, const char *fmt, ...)
     va_list argptr;
     static char data[4096];
     FILE *fp;
```
TeslaRus: oh, I forget to rewrite it using SDL_rwops...

PR #353, opened 2016-06-30
Camera/crosshair small refactoring + inclusion fix

Banderi:
Separated camera functions into `game_camera.cpp`;
Added toggle-able debug crosshair;
Added inclusion to <ctype.h> in `gui.cpp` (for `tolower()`);


src/gui.cpp
```diff
@@ -246,6 +246,10 @@ void Gui_Render()
     qglPixelStorei(GL_UNPACK_LSB_FIRST, GL_FALSE);
     qglPixelStorei(GL_UNPACK_ALIGNMENT, 1);
 
+    if (screen_info.show_debuginfo != 0)
```
TeslaRus: here should be something like 'if(screen_info.show_crosshair)'; compare crosshair and debug info is a hack;

src/gui.cpp
```diff
@@ -924,7 +928,7 @@ void Gui_FillBackgroundBuffer()
     qglBufferDataARB(GL_ARRAY_BUFFER, sizeof(GLfloat[32]), backgroundArray, GL_STATIC_DRAW);
 }
 
-void Gui_DrawCrosshair()
```
TeslaRus: It is better to stay Gui_DrawCrosshair() function;

src/gui.cpp
```diff
@@ -934,6 +938,12 @@ void Gui_DrawCrosshair()
     qglDrawArrays(GL_LINES, 0, 4);
 }
 
+void Gui_DrawCrosshair()
```
TeslaRus: Gui_DrawCrosshair() should be restored;

src/game_camera.cpp
```diff
@@ -0,0 +1,302 @@
+
+#include <stdlib.h>
```
TeslaRus: I correct understand: these code was mowed from camera without changes?

src/gui.cpp
TeslaRus: if(screen_info.show_debuginfo != 0) // by used code style 'if' and '(' goes without space

src/gui.cpp
Banderi: If we have Gui_DrawDebugCrosshair(), we can use Gui_DrawCrosshair() to draw the actual ingame gameplay crosshair when it'll be implemented, instead of a using a temporary opengl buffer.

src/game_camera.cpp
Banderi: Yes, the Cam_ functions do not depend on these inclusions.

src/game_camera.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+#include "core/system.h"
+#include "core/obb.h"
+#include "render/camera.h"
+#include "controls.h"
+#include "room.h"
+#include "world.h"
+#include "skeletal_model.h"
+#include "entity.h"
+//#include "script.h"
```
TeslaRus: commented includes should be deleted

src/gui.cpp
TeslaRus: maybe, but stay here unimplemented empty function is not good too; this time will be used setting flag from config, so if crosshair does not needed, it could be disabled by config (+later by console command); when new crosshair be implemented, then Gui_DrawCrosshair() will be updated; do not needed for additional functions creation;

src/game_camera.cpp
Banderi: I left those three as I though they might be used in conjunction with the current code to implement correct camera controls in the future. I can remove them if it's bad practice.

src/gui.cpp
Banderi: Sounds fair, I'll re-merge them.

PR #355, opened 2016-07-02
Anim control interface

TeslaRus:
Updated anim control functional (and script) interface: added multi animation support;

### DIFF
--- a/src/entity.cpp
+++ b/src/entity.cpp
@@ -951,7 +951,7 @@ void Entity_SetAnimation(entity_p entity, int anim_type, int animation, int fram
             entity->anim_linear_speed = entity->bf->animations.model->animations[animation].speed_x;
         }
         SSBoneFrame_SetAnimation(entity->bf, anim_type, animation, frame);
-        SSBoneFrame_Update(entity->bf);
+        SSBoneFrame_Update(entity->bf, 0.0f);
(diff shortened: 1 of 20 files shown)


src/entity.cpp
```diff
@@ -937,21 +938,22 @@ void Entity_ProcessSector(entity_p ent)
...
-    entity->anim_linear_speed = entity->bf->animations.model->animations[animation].speed_x;
-    SSBoneFrame_SetAnimation(entity->bf, animation, frame);
-    SSBoneFrame_Update(entity->bf);
-    Entity_UpdateRigidBody(entity, 0);
+        if(anim_type == ANIM_TYPE_BASE)
+        {
+            entity->anim_linear_speed = entity->bf->animations.model->animations[animation].speed_x;
+        }
+        SSBoneFrame_SetAnimation(entity->bf, anim_type, animation, frame);
+        SSBoneFrame_Update(entity->bf);
```
TeslaRus: that function call should not be here;

src/entity.cpp
```diff
@@ -937,21 +938,22 @@ void Entity_ProcessSector(entity_p ent)
...
+        Entity_UpdateRigidBody(entity, 0);
```
TeslaRus: that function call should not be here;

Issue #94, opened 2015-06-24
Wrong initial trigger state


vvs-:
In *Atlantis* pillars in room 58 are in the wrong order. If level is loaded in noclip state then they have correct order.

Lwmte:
I remember a bug with flipped rooms which incorrectly set invisible entity attribute to visible (e.g. falling ceiling in Tomb of Qualopec). Possibly whole entity structure is messed up when rooms are flipped.

Lwmte:
Is it still happening?

vvs-:
Yes. Just try *Atlantis* level. No doors are opening in the initial room until you flip it manually.

vvs-:
Seems that the reason for this is simple: flipping doesn't work from initial position. I can only flip that level when Lara has moved from the trigger sector.

vvs-:
The _Atlantis_ initial flip bug happens for several reasons. First, there is a one-shot flag which prevents trigger from working twice. And there is no any entity linked to this trigger at all, just FLIPMAP. Second, it uses flip effect 16, which flips rooms several times back and forth with predefined delay.

TeslaRus:
TR1_MAP14 start: trigger: once = true; **cmd1**: FlipMap[0].mask = 0x1F; flipped to alt; **cmd2**: flicker;
Questions: 
flicker works with flipmap[0] hardcoded?
may be it only flip rooms to back without flipmap routine?

vvs-:
That's a good question. I tested FLICKER_FX on a test map, but didn't try to implement several FLIPMAPs there.

Gh0stBlade:
Can someone please confirm what a "FLIP MAP" is?

vvs-:
FLIPMAP is a trigger command which change flip state for alternate rooms. It's described in TRS3.

Issue #33, opened 2015-05-17
Wrong lever behavior


vvs-:
In *St. Francis' folly* lever 66 can be operated only once, while in the original it's unlimited. That leads to possible deadlock if player didn't trigger closing a door when entered the room first time.

Lwmte:
I noticed that when testing this particular switch setup. Strange thing is that in original level, switch trigger for switch 66 is actually set as one-shot, which leads me to a confusion, because one-shot state flag SHOULD indeed lock the switch after first use.

So either it is hardcoded override (as with TR2 spinning blades sound override in Great Wall), or TR1 switches are processed a bit differently and won't lock up on first use. Actually, I remember something like that, I was able to un-pull certain one-shot switches in TR1.

Just another headache, thanks to Core Design.

vvs-:
That gave me an idea: why not implement a run-time patching for original levels? I've seen other bugs in them as well, so it's the only way to fix them all. Some other FLOSS engines already do just that, if I'm not mistaken that's *Exult*.

Gh0stblade:
I thought LUA scripts could override triggers already?

vvs-:
Yes, they could by they don't at this time.

Lwmte:
I should add an option to print out all auto-generated trigger functions, and then an option to override certain trigger function with user-specified one. I just afraid that if it's common option in TR1 to have one-shot switches work two-way, it's better to make automated as well.

vvs-:
Or you could patch a trigger even before generating Lua code like static meshes.

vvs-:
It might or might not be related. In *The Cistern* room 10, switch 11 works only once too. It can be pulled multiple times, but only the first one flips the rooms outside. So, you can fill these rooms with water, but can't flush it.

vvs-:
I think the entity `lock` has incorrect behavior. Right now it locks the entity and prevents updates to its' mask. Actually, it should lock the trigger from changing its' state again and not the entity properties.

Right now it prevents *Temple of the Cat* from being completed. Any one of the two triggers locks the entity and prevents another trigger from activating it.

Lwmte:
The problem is, original TRs have *no such thing* as trigger state. That is, every time you step onto trigger, it is executed again and again. Only entities have lock state assigned to them. So basically, when trigger is set as *one-shot*, it basically means that any entity which was activated using this trigger, won't be activated again, ever.

It's just TR1 which presumably has different trigger/entity processing, and we still need to understand the difference.

[...]

PR #361, opened 2016-07-18
Triggers reorganization

TeslaRus:
added entity activation / deactivation ret code using;


vvs-:
In case of pre-activated entities the states are inverted. _Activated_ means mask 0 and _deactivated_ means mask 31. So, it should work other way round.

And there is another bug discussed [here](https://github.com/opentomb/OpenTomb/issues/33#issuecomment-125575819). It was fixed in `backup` branch.

vvs-:
Yet another difference. Entity timer is active only when will be locked and the action ignores the lock. Also, when entity will not be locked the timer can be activated by just the trigger pad. Otherwise the trigger pad performs just the action. The switch timer always works independent.

vvs-:
OT ignores anything after the first item in action list of HEAVY trigger.

TeslaRus:
> OT ignores anything after the first item in action list of HEAVY trigger.
Looks like (trigger.cpp string 313)
```
if(!IsPlayer && (entity_activator->trigger_layout & ENTITY_TLAYOUT_SSTATUS))
{
    return;
}
```

vvs-:
Bingo. That's the culprit.

vvs-:
DUMMY trigger should not act on non-activated trapdoors.

vvs-:
SWITCH trigger works for FLIPMAP just the same as for OBJECT, i.e. it should XOR the mask.

vvs-:
As #94 shows FLIPMAP has a lock flag and it can be locked. Confirmed by a test map.

vvs-:
There is a new bug in heavy triggers: FLIPMAP doesn't work. It means that _Palace Midas_ don't work too.

vvs-:
In OpenTomb's present implementation using setFlipMap(0,0x1F,TRIGGER_OP_XOR) isn't sufficient to flip rooms when the mask becomes zero. We need to further investigate how it's implemented in the original TR1.

vvs-:
Further investigation shows that ANTI-PAD doesn't deactivate FLIPMAP, but SWITCH does. So, the zeroed mask is ignored by FLIP-MAP/ON/OFF, but SWITCH trigger has its own implementation. It doesn't actually invert the flip state, but always activates and deactivates alternate rooms when the switch activates and deactivates.

vvs-:
And contrary to what TRS said ANTI-PAD _does_ indeed use the mask.

vvs-:
Collapsible floor, trapdoors and perhaps some other objects has no collision by default. Any trigger with OBJECT command which points to it applies collision when activated. DUMMY trigger is usually used for that purpose, but if it points to collapsible floor that floor tile doesn't collapse and is solid. Only after that tile is activated by some other trigger it would collapse.

TeslaRus:
fixed trigger PAD in setgamef(1, 1) room 17

vvs-:
But there's still one problem remains. The door doesn't remember the last trigger mask while in the original TR1 it does. You can test it by quickly rolling over triggers during state change and jumping from it before it finished.

TeslaRus:
entity.cpp #1180 str: Entity_Activate(...) function: mask set to entity if current timer == 0 and entity are not locked. Trigger mask may be dropped by calling Entity_Deactivate(...) (entity.cpp #1242 str). In my case door closes immediately after opening complete... but it is because timer was started after beginning opening and finish in ending of opening. Oh... I have to add layout flags to trigger debug info!

TeslaRus:
Added layout info to trigger debug + trigger debug info may be showed on marked sector by RMB (right mouse button), not only sector with Lara;

vvs-:
This is bad test case because it has only one trigger. Use PAD/ANTI-PAD from Lwmte's test level. If you manage to jump from trigger sector before animation is complete the door state won't correspond to the last active trigger. In TR1 it will always change it state even when the last trigger is no longer active. That means that the door mask is always changed during animation and the door will check it after animation was complete.

vvs-:
Also, there is a crash now inside ShowDebugInfo:
```
#0  ShowDebugInfo () at src/engine.cpp:902
#1  0x08097725 in Engine_Display () at src/engine.cpp:513
#2  0x080978bd in Engine_Display () at src/engine.cpp:776
#3  Engine_MainLoop () at src/engine.cpp:791
#4  0x08085c43 in main (argc=1, argv=0xbffff394) at src/main_SDL.cpp:18
```

TeslaRus:
Crash was fixed

src/world.cpp
```diff
@@ -972,7 +972,7 @@ int World_SetFlipState(uint32_t flip_index, uint32_t flip_state)
         return 0;
     }
 
-    if(global_world.flip_map[flip_index] == 0x1F)         // Check flipmap state.
+    if((global_world.flip_map[flip_index] == 0x1F) || (flip_state == 0))        // Check flipmap state.
```
vvs-: Switch trigger use flip_state=1 presently, so it doesn't work. FLIP_OFF is the only trigger that use it in OT.

TeslaRus:
but logic here http://www.pasteall.org/72318/c says that if (flip_map[flip_index] != 0x1F) - than flip rooms back

vvs-:
Actually it always checks that mask is 0x1F:
```
case FLIP_OFF:
    if((flipMasks[actionParam] & 0x3e00) == 0x3e00 && roomsAreSwapped)
```

TeslaRus:
I reread code and saw mistake: fixed ;-)

vvs-:
Yes, it works correctly now. The base room has a pillar while the alternate don't. There are two triggers in the first room: a PAD and a SWITCH. Both have FLIPMAP command.

vvs-:
Yet another bug: the text in inventory screen is not centered.

TeslaRus:
fixed (text coordinates vas updated for new alignment logic)

vvs-:
Yet another bug: HEAVY trigger activates FLIPMAP every tick and not just once. This breaks _Palace Midas_ again.

src/world.cpp
```diff
@@ -2208,7 +2208,7 @@ void World_GenRoom(struct room_s *room, class VT_Level *tr)
     room->alternate_room = NULL;
     room->base_room = NULL;
 
-    if((tr_room->alternate_room >= 0) && ((uint32_t)tr_room->alternate_room < tr->rooms_count))
+    if((tr_room->alternate_room >= 0) && ((uint32_t)tr_room->alternate_room < tr->rooms_count) && (room->id < tr_room->alternate_room))
```
vvs-: The original engine has more complex logic. The one here doesn't work in all circumstances. We need to find out how the original algorithm works.

vvs-:
There's now something wrong with the camera in the _Palace Midas_. It's blinking when Lara is pulled by a UWCURRENT.

src/world.cpp
stohrendorf: It's basically a `swap(rooms[roomId], rooms[rooms[roomId].alternateRoom])`, at least for TR1, meaning you can actually _cycle_ through rooms by chaining them through different `alternateRoom` ids.

vvs-: And how it works in case when `alternateRoom` is -1?

stohrendorf: As said, TR1 swaps the _contents_ of the room structure, not pointers. It does not use any "base room" reference or such; it only swaps the contents (i.e., all items that can have a room ownership) using the alternative room id; if it is `<0`, no swapping occurs.

Example: If room 0 has alternative=1, and room 1 has alternative=2, (i.e. [1,2,...]) on the first swap all items are removed from room 0 and moved to room 1 (discarding all items from room 1), then the contents of the structures are swapped, so the data occurs as if room 1 has been the "base room" from the beginning. After the swap, the alternative ids are now [2, -1,...], because the alternative room's alternative room id (yikes!) is set to -1.

vvs-: But how TR1 determined which room to begin with? For example it can start in room 1 and then swap to room 2. But at the same time there are examples when it starts in room 2 and swaps to room 1. How it finds out which room should be the first?

vvs-: Now I have a test map with three rooms. First room has the second as alternate, that has the third as alternate which has alternate room set to -1. All rooms has the same FLIPMAP, FLIPON and FLIPOFF triggers. But I can't cycle through these rooms, because the second alternate room get ignored and it swapped back to the first room.

stohrendorf: Yeah, maybe some example with the alternative ids [0..4].
[0, 1, 2, 3, 4] -> swap ->[1, -1, 3, -1, -1]

Basically:
```
foreach(room1 in rooms) { // in ascending order
  if(room1->alternative < 0) continue;
  room2 = &rooms[room1->alternative];
  // move items/blocks/etc. from room1 to room2, discarding every dynamic item in room2
  swap(*room1, *room2); // swap everything except the 'alternative' flag
  swap(room1->alternative, room2->alternative);
  room2->alternative = -1;
}
```

vvs-: May be I didn't understand it correctly, but using the code:
```
#include <stdio.h>

void swap(int *a, int *b) {
	int t;
	t = *a;
	*a = *b;
	*b = t;
	return;
}

int main() {
  int rooms[] = {1,2,3,4}, room1, room2;
  for (room1 = 0; room1 < sizeof(rooms)/sizeof(int)-1; room1++) {
	  if (rooms[room1] < 0) continue;
	  room2 = rooms[room1];
	  swap(&rooms[room1], &rooms[room2]);
	  rooms[room2] = -1;
  };
  printf("[%d,%d,%d,%d]\n",rooms[0],rooms[1],rooms[2],rooms[3]);
  return 0;
}
```
I got `[2,-1,4,-1]`.

I found how the engine determined a starting room. It just used PORTAL function. It also used initial Lara position without transferring her into another room.

stohrendorf: You are basically only swapping room->alternativeRoom, but you need to swap the rooms _and_ then the alternative IDs; as your pseudo-room only consists of the alternative room ID, you need to add another `swap(&rooms[room1], &rooms[room2])`. Maybe better use a struct with 2 members (an id for debugging the room data swap, and the alternative room id) in your case.

vvs-:
Camera doesn't pass heavy1 test. Switch triggers doesn't work.

TeslaRus:
I mess true and false in flag initialization, so switches was broken... perfect error...
This time camera == heavy trigger still not implemented (but engine has info about usage camera as a heavy trigger);

vvs-:
Seems like camera jerking is caused by two conflicting functions. One is trying to follow Lara and other is trying to center on trigger TARGET. This is visible at the beginning of _Palace Midas_ underwater scene.

PR #362, opened 2016-07-19
[NOT_FOR_MERGING] Altroom test

vvs-:
This is a level I made to test room flip triggers. It works on TR1 but for some reason crashes OT. I suspect that there are some issues with portals between rooms. This level was created by DXtre3D.

I'd like to find the reason for this incompatibility. Could someone look at it?


vvs-:
Never mind, found the reason.

This patch fixes it:
```
diff --git a/src/world.cpp b/src/world.cpp
index b653d7e..5ce5246 100644
--- a/src/world.cpp
+++ b/src/world.cpp
@@ -2208,7 +2208,7 @@ void World_GenRoom(struct room_s *room, class VT_Level *tr)
     room->alternate_room = NULL;
     room->base_room = NULL;
 
-    if((tr_room->alternate_room >= 0) && ((uint32_t)tr_room->alternate_room < tr->rooms_count))
+    if((tr_room->alternate_room >= 0) && ((uint32_t)tr_room->alternate_room < tr->rooms_count) && (room->id < tr_room->alternate_room))
     {
         room->alternate_room = global_world.rooms + tr_room->alternate_room;
     }
```

PR #364, opened 2016-07-25
Add test level for alternate room with PORTAL sector

vvs-:
This test level demonstrate how engine should process portal sectors. In this level this portal leads to room 2. OpenTomb incorrectly points to room 1 instead.

Another test level demonstrates rooms cycling. Many thanks goes to @stohrendorf for the idea and reverse engineering!


vvs-:
Nah... Where is your curiosity? And where is a mystery? I'm heartbroken :sob:

PR #365, opened 2016-07-28
Triggers reorganization

TeslaRus:
second stage of triggers reorganization;


TeslaRus:
Useful discussion information from first stage:
 stohrendorf added a note 4 days ago

It's basically a swap(rooms[roomId], rooms[rooms[roomId].alternateRoom]), at least for TR1, meaning you can actually cycle through rooms by chaining them through different alternateRoom ids.
@vvs-
OpenTomb member
vvs- added a note 4 days ago • edited

And how it works in case when alternateRoom is -1?

Actually, I was interested in what room is considered alternate and which is base. Right now OT determines it by -1 in alternate_room pointer which is a weak test.

I suppose something like this should work, but still interested in the original implementation:

diff --git a/src/world.cpp b/src/world.cpp
index 6ca54b5..188b276 100644
--- a/src/world.cpp
+++ b/src/world.cpp
@@ -2210,7 +2210,7 @@ void World_GenRoom(struct room_s *room, class VT_Level *tr)
     room->alternate_room = NULL;
     room->base_room = NULL;

-    if((tr_room->alternate_room >= 0) && ((uint32_t)tr_room->alternate_room < tr->rooms_count) && (room->id < tr_room->alternate_room))
+    if((tr_room->alternate_room >= 0) && ((uint32_t)tr_room->alternate_room < tr->rooms_count) && (room->id < tr_room->alternate_room || (global_world.rooms + tr_room->alternate_room)->alternate_room != room))
     {
         room->alternate_room = global_world.rooms + tr_room->alternate_room;
     }

@stohrendorf
OpenTomb member
stohrendorf added a note 4 days ago

As said, TR1 swaps the contents of the room structure, not pointers. It does not use any "base room" reference or such; it only swaps the contents (i.e., all items that can have a room ownership) using the alternative room id; if it is <0, no swapping occurs.

Example: If room 0 has alternative=1, and room 1 has alternative=2, (i.e. [1,2,...]) on the first swap all items are removed from room 0 and moved to room 1 (discarding all items from room 1), then the contents of the structures are swapped, so the data occurs as if room 1 has been the "base room" from the beginning. After the swap, the alternative ids are now [2, -1,...], because the alternative room's alternative room id (yikes!) is set to -1.
@vvs-
OpenTomb member
vvs- added a note 3 days ago • edited

But how TR1 determined which room to begin with? For example it can start in room 1 and then swap to room 2. But at the same time there are examples when it starts in room 2 and swaps to room 1. How it finds out which room should be the first?

Yet another point. All alternate rooms created by Core editor has alternate room id set to -1. According to what you've said it can't swap from alternate room back to the base/starting room in such cases. There should be some other rules.
@vvs-
OpenTomb member
vvs- added a note 3 days ago • edited

Now I have a test map with three rooms. First room has the second as alternate, that has the third as alternate which has alternate room set to -1. All rooms has the same FLIPMAP, FLIPON and FLIPOFF triggers. But I can't cycle through these rooms, because the second's alternate room get ignored and it swapped back to the first room.
@stohrendorf
OpenTomb member
stohrendorf added a note 3 days ago • edited

Yeah, maybe some example with the alternative ids [1..4] and [1..5].

[1, 2, 3, 4]    -> swap ->[ 1, -1, 3, -1]
[1, 2, 3, 4, 0] -> swap ->[-1, -1, 3, -1, 0]

Basically:

foreach(room1 in rooms) { // in ascending order
  if(room1->alternative < 0) continue;
  room2 = &rooms[room1->alternative];
  // move items/blocks/etc. from room1 to room2, discarding every dynamic item in room2
  swap(*room1, *room2); // swap everything except the 'alternative' flag
  swap(room1->alternative, room2->alternative);
  room2->alternative = -1;
}

@vvs-
OpenTomb member
vvs- added a note 3 days ago

May be I didn't understand it correctly, but using the code:

#include <stdio.h>

void swap(int *a, int *b) {
    int t;
    t = *a;
    *a = *b;
    *b = t;
    return;
}

int main() {
  int rooms[] = {1,2,3,4}, room1, room2;
  for (room1 = 0; room1 < sizeof(rooms)/sizeof(int)-1; room1++) {
      if (rooms[room1] < 0) continue;
      room2 = rooms[room1];
      swap(&rooms[room1], &rooms[room2]);
      rooms[room2] = -1;
  };
  printf("[%d,%d,%d,%d]\n",rooms[0],rooms[1],rooms[2],rooms[3]);
  return 0;
}

I got [2,-1,4,-1].

I found how the engine determined a starting room. It just used PORTAL function. It also used initial Lara position without transferring her into another room.
@stohrendorf
OpenTomb member
stohrendorf added a note 3 days ago

You are basically only swapping room->alternativeRoom, but you need to swap the rooms and then the alternative IDs; as your pseudo-room only consists of the alternative room ID, you need to add another swap(&rooms[room1], &rooms[room2]). Maybe better use a struct with 2 members (an id for debugging the room data swap, and the alternative room id) in your case.
@vvs-
OpenTomb member
vvs- commented 4 days ago • edited

There's now something wrong with the camera in Palace Midas. It's blinking when Lara is pulled by a UWCURRENT.
vvs- and others added some commits 3 days ago
	@vvs- 	Add test level for alternate room with PORTAL sector 		
	929caa2
	@TeslaRus 	fixed trigger header conditions;
		
	8448daf
	@vvs- 	Add cycle rooms test level 		
	1276625
@vvs-
OpenTomb member
vvs- commented 3 days ago

Camera doesn't pass heavy1 test. Switch triggers doesn't work.
@stohrendorf
OpenTomb member
stohrendorf commented on 1276625 2 days ago

Test case description is missing. What should be done to test this, what is the expected result?
vvs- and others added some commits 2 days ago
	@vvs- 	Add a short description 		
	ee40c5a
	@TeslaRus 	fixed switch logick (messed true and fals in initialization... I need…
		
	a2e3e7b
	@TeslaRus 	Merge pull request #364 from opentomb/portal_test
		
	00e226d
@TeslaRus
OpenTomb member
TeslaRus commented 2 days ago

I mess true and false in flag initialization, so switches was broken... perfect error...
This time camera == heavy trigger still not implemented (but engine has info about usage camera as a heavy trigger);
@vvs-
OpenTomb member
vvs- commented a day ago

Seems like camera jerking is caused by two conflicting functions. One is trying to follow Lara and other is trying to center on trigger TARGET. This is visible at the beginning of Palace Midas underwater scene.

vvs-:
Also, these differences are still need to be fixed:
- [ ] https://github.com/opentomb/OpenTomb/pull/361#issuecomment-233310995
- [ ] https://github.com/opentomb/OpenTomb/pull/361#issuecomment-233330148
- [ ] https://github.com/opentomb/OpenTomb/pull/361#issuecomment-233520980
- [ ] https://github.com/opentomb/OpenTomb/pull/361#issuecomment-233656973
- [ ] https://github.com/opentomb/OpenTomb/pull/361#issuecomment-234180606
- [ ] https://github.com/opentomb/OpenTomb/pull/361#issuecomment-234309192
- [ ] https://github.com/opentomb/OpenTomb/pull/361#issuecomment-234530637

Gh0stBlade:
Ok so I assume this is related to the new room flip bugs.

I have noticed a couple in TR3's Temple of Puna:

1. When no clipping out of the world, alternate room states are shown incorrectly.
2. Room 22 has serious issues, pulling the switch does not flip the room?
3. After pulling the switch, leaving the room with no-clip and entering it, the correct room is displayed. Except the room collision seems to be overlapping with it's original room. This leads me to believe that both rooms are active at the same time.
4. Entering Room 23 is extremely buggy, Lara is pushed by something invisible.

vvs-:
Actually, nobody even tested triggers yet on anything newer than TR1. And there are significant differences in implementation of TR3+.

PR #372, opened 2016-08-02
Merge.

Gh0stBlade:


src/render/frustum.cpp
```diff
@@ -22,6 +20,7 @@ CFrustumManager::CFrustumManager(uint32_t buffer_size)
     m_buffer_size = buffer_size;
     m_allocated = 0;
     m_buffer = (uint8_t*)malloc(buffer_size * sizeof(uint8_t));
+    memset(m_buffer, 0, (buffer_size * sizeof(uint8_t)));
```
TeslaRus: here can be garbage in buffer, because all used data are overridden by algorithms

PR #373, opened 2016-08-02
Stability fixes.

Gh0stBlade:


src/world.cpp
```diff
@@ -2521,6 +2521,11 @@ void World_GenRoomCollision()
 {
     room_p r = global_world.rooms;
 
+    if(r == NULL)
+    {
+        return;
```
TeslaRus: for will not iterate if no rooms in world (count check)

src/world.cpp
```diff
@@ -2589,6 +2600,11 @@ void World_MakeEntityItems(struct RedBlackNode_s *n)
 {
     base_item_p item = (base_item_p)n->data;
 
+    if(item == NULL)
+    {
+        return;
```
TeslaRus: it is optimization =)

src/world.cpp
Gh0stBlade: Correct, except i feel that, it's good practice to check all pointers against NULL. Either way, if you have the room count set to 1 and global_world.rooms is actually null. That will cause a lot of problems, not saying it will happen but it's always a possibility. Perhaps it would have been better to use an assertion. I can revert this if you wish, let me know :)

stohrendorf: `assert(rooms == NULL || rooms_count > 0);`?

TeslaRus: if(r == NULL) then return - is normal condition;
assert(rooms == NULL || rooms_count > 0); - there are cases engine try to load level that not exists, so runtime checks is needed here + some original levels I load contains only textures (TR5)

PR #380, opened 2016-09-05
Update npc_behaviour.asc

XProger:
In TR1, bit 15 of tr_box.OverlapIndex must be skipped to get real index in the overlap array. bit-15 is unknown, it may occur behind a door, I've no idea for what... yet

### DIFF
--- a/doc/npc_behaviour.asc
+++ b/doc/npc_behaviour.asc
@@ -79,7 +79,7 @@ struct tr_box   // 20 bytes
     uint32_t Xmin;
     uint32_t Xmax;
      int16_t TrueFloor;     // Height value in global units
-     int16_t OverlapIndex;  // Index into Overlaps[].
+     int16_t OverlapIndex;  // bit 0..14 is the index into Overlaps[], bit 15 is unknown (may occur behind a door)
 };
 ----
 
@@ -93,7 +93,7 @@ struct tr2_box   // 8 bytes
     uint8_t Xmin;
     uint8_t Xmax;
     int16_t TrueFloor;     // Height value in global units
-    int16_t OverlapIndex;  // Index into Overlaps[]. 
+    int16_t OverlapIndex;  // Index into Overlaps[] (asme as tr_box?) 
 };
 ----
 
--- a/doc/npc_behaviour.asc
+++ b/doc/npc_behaviour.asc
@@ -93,7 +93,7 @@ struct tr2_box   // 8 bytes
     uint8_t Xmin;
     uint8_t Xmax;
     int16_t TrueFloor;     // Height value in global units
-    int16_t OverlapIndex;  // Index into Overlaps[] (asme as tr_box?) 
+    int16_t OverlapIndex;  // Index into Overlaps[] (same as tr_box?) 
 };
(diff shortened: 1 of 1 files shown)


doc/npc_behaviour.asc
```diff
@@ -79,7 +79,7 @@ struct tr_box   // 20 bytes
     uint32_t Xmin;
     uint32_t Xmax;
      int16_t TrueFloor;     // Height value in global units
-     int16_t OverlapIndex;  // Index into Overlaps[].
+     int16_t OverlapIndex;  // bit 0..14 is the index into Overlaps[], bit 15 is unknown (may occur behind a door)
```
stohrendorf: Could you change the type to `uint16_t` to make sure that using `abs()` is the wrong way of masking bit 15 out?

Gh0stBlade:
I looked at the Overlaps August last year. Bit 15 is a flag used to terminate iterating over the overlap list in the original engine. Basically, ‭(OverlapIndex & 0x8000‬) = 1 means it's the last overlap to process in a "list".

1. A box is pretty much a sector:

----------------
|                     |
|       BOX       |
|                     |
----------------
 
The Box struct contains an index into the overlaps list.
Example: Overlap* o = &world_overlaps[box->overlap_index];

===================================================

2. Overlap struct also points to a single box . This way, we can access a box's overlaps' box by:

while(!(o->box_index&0x8000))
{
   Box* b = &world_boxes[o->box_index&0x7FFF];//This is the box the current overlap points to.
   o++;
}

------------------------------------------------
|                     |                      |                       |
|    Overlap    |      Overlap    |     Overlap     |
|                     |                      |                       |
------------------------------------------------
|                     |                      |                      |
|    Overlap    |        Box        |     Overlap     |
|                     |                      |                      |
------------------------------------------------
|                     |                      |                      |
|    Overlap    |      Overlap    |     Overlap    |
|                     |                      |                      |
------------------------------------------------

-----------------------------------------------------------------------------------
The process basically allows us to check all overlaps of a single box without:
1. Having to check if an overlap/box/sector exists (cases where Lara is close to walls).
2. Having to pre-compute lists of neighboring sectors/boxes/overlaps at run-time.
3. It's limited since using the pre-computed data would not allow us to have control over certain sectors. Some sectors have no box/overlap index so AI entites can't move there (hence why I think A* is a better option).

Cheers.

XProger:
I think, it should be something like:
`struct Overlap {
     uint16 box_index:15, end:1;
} *o = &world_overlaps[box->overlap_index & 0x7FFF]; // overlap_index can be 0xFFFF ?
do {
    Box* b = &world_boxes[o->box_index]; // boxIndex can be 0xFFFF ?
    // ...
} while (!(o++)->end);`
but I'll do research

XProger:
these overlaps not only used by enemies but Lara's too
image
as you see ceiling (red) & floor (green portals) at the perimeter of pool have randomly height. I think, for get out of the pool I need to use box overlap info.
image

Gh0stBlade:
Yep you're right. I'm more awake now. My comment has been updated to prevent confusion.

PR #381, opened 2016-09-10
Update meshes_models.asc

XProger:
fix static mesh bounding box description by adding tr_bounding_box structure (it's not just two vectors)

### DIFF
--- a/doc/meshes_models.asc
+++ b/doc/meshes_models.asc
@@ -136,7 +136,7 @@ struct tr_bounding_box // 12 bytes
 };
 ----
 
-`VisibilityBox` and `CollisionBox` boundaries is always stay axis aligned even after applying tr_room_staticmesh::Rotation (always have 90 degrees step)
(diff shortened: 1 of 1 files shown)


doc/meshes_models.asc
```diff
@@ -123,16 +123,20 @@ and the latter is used for collision testing.
...
+
+struct tr_bounding_box // 12 bytes
+{
+    int16_t MinX, MaxX, MinY, MaxY, MinZ, MaxZ; 
 };
 ----
 
-For `VisibilityBox` and `CollisionBox`, dimensions are specified using two vectors, first being the `bounding box minimum`, and second being the `bounding box
-maximum`.
+`VisibilityBox` and `CollisionBox` boundaries is always stay axis aligned even after applying tr_room_staticmesh::Rotation (always have 90 degrees step)
```
stohrendorf: Additionally, the test whether to rotate the box or not relies on the mesh's rotation being and exact multiple of `0x4000` (aka 90 degrees). If this is not the case, the box is not rotated, which results in wrong collision checks.

stohrendorf: @XProger If you could add the comment above to an appropriate place, this merge request can be merged.

Issue #343, opened 2016-06-26
Incorrect transfer of entities in room flipping


vvs-:
There are multiple models with wrong collision flags. Models 61 and 64 in _Tomb of Qualopec_, model 63 in _Atlantis_, model 48 and static 37 in _Tomb of Tihocan_, model 48 in _City of Khamoon_. But the same model can have different flags in the same level, e.g. entities 1 and 28 in _City of Khamoon_. And the most exciting is that if you teleport there from the beginning of the level the flags will be different than if you go there by normal means or even noclip.

vvs-:
I found that it's because room flipping is buggy.

vvs-:
Key 38 is missing in _City of Khamoon_ as well.

vvs-:
Also, entity's trigger layout is not updated in alternate room.

vvs-:
In original engine all doors are transferred deactivated, but switches remained intact. In OT everything is transferred as is.

vvs-:
Keyholes doesn't transfer in OT as well as keys.

Issue #259, opened 2015-08-09
Excessive fuse


vvs-:
There is an excessive fuse in _Natla's Mines_ before a conveyor belt. It shouldn't be there until a conveyor is turned on.

Lwmte:
I believe the reason is we don't currently have functionality for *invisible* item flag. This flag, along with *clear body* flag, was kept in originals along with activation mask and only once flag in same 8-bit bitfield.

vvs-:
Perhaps, it is related to #93 then?

Lwmte:
Oh yes, you are perfectly right! Boat has *invisible* flag set, as I checked it with FexInspect.

Have you just guessed it, or analyzed flag structure for yourself? I'm asking because I still confuse "clear body" and "invisible" flag placement.

vvs-:
I never used any unofficial tools. First of all I don't use Windows. There is everything I need in OpenTomb, except editor :disappointed:

Lwmte:
You can always use Wine! ;)
FexInspect is really cool utility. You can easily analyze level structure with it, including triggers, FloorDara layout, entity positions, light positions, etc. I'd highly recommend it to anyone who's trying to debug original levels.

FexInspect and Wadmerger are two main utilities which I use to analyze original TR data. Without them, the progress on engine development would be MUCH slower.

vvs-:
> You can always use Wine! ;)

Well, it's fat and buggy.

Nevertheless, I think you are right. I'll eventually try to setup Wine for this utilities and most important for editor.

vvs-:
Also, pistols in room 94 have invisible flag set.

vvs-:
There are many invisible entities in TR1, e.g. falling ceiling or movable blocks over Thor room. After implementing invisible flag parsing these movable blocks are not activated in Lua script and remain invisible.

PR #384, opened 2016-09-25
Alt room refactoring

TeslaRus:
alt rooms used only for content store;
no more "check flip" functions: only real_room ptr or Sector_CheckRealRoom(...);
Room collision generation is workaround, because real tween collision between rooms depend on room combination (alt / real version) so there are two ways: 1) generate additional tween collision meshes for ALL room / real room combinations; 2) generate on fly after every World_SetFlipState();
UPD: only 2) is sane solution;


vvs-:
In _Tomb of Qualopec_ portal from room 18 to room 11 doesn't work, Lara can't pass.

In _Cistern_ swapping to water room doesn't transfer its water attribute, only the water surface is seen.

#343 is still valid.

In altroom4 test level Lara starts in a wrong room.

TeslaRus:
"In altroom1 test level room swapping doesn't work." - I test it, it does not works because I stay as is swapping sequence (if I swap 3 cycled rooms from 1 to 3, then 1 made full turn and no changes for 1 (real room), 2 and 3 changed). How to correctly swap rooms? does I need to mark **next_alt_room** as swapped / unswapped (like world.cpp str 976)?

TeslaRus:
> In Cistern swapping to water room doesn't transfer its water attribute, only the water surface is seen.
Thx, forgot flags

TeslaRus:
about altroom initial pickable objects state: in Natla's mines pistols and large med is accessible before 3 puzzle hole  activation (or it is an additional script with objects enabling / disabling)?

vvs-:
> 1 made full turn and no changes for 1 (real room), 2 and 3 changed)
Are you sure it's altroom1? There is only two that should be swapped. One has a raised cube and the other doesn't.

TeslaRus:
if rooms are cycled, swapping does twice, so no visible changes

TeslaRus:
altroom1 passed;
room_flags fixed;
physics crash fixed;
World_FindRoomByPosCogerrence(...) fixed;

vvs-:
There is something wrong with the water rooms in _Cistern_. After swapping rooms, Lara can still swim in the air after climbing from water.

There are no changes in _Tomb of Qualopec_ and altroom4 also.

vvs-:
> about altroom initial pickable objects state: in Natla's mines pistols and large med is accessible before 3 puzzle hole activation (or it is an additional script with objects enabling / disabling)?

I'm not sure. But in TR3 NPC inventory is kept under their feet. Maybe that's the same?

vvs-:
> about altroom initial pickable objects state: in Natla's mines pistols and large med is accessible before 3 puzzle hole activation (or it is an additional script with objects enabling / disabling)?

This is actually described in #259.

vvs-:
In _Natla's Mines_ camera repeatedly bumps into invisible boat static near waterfall.

vvs-:
In _Natla's Mines_ there is something wrong with sector `4,1` in room 28. Lara consistently lose her room ownership and falls out, sometimes into room 20 and sometimes out of level mesh.

vvs-:
altroom2 test was (and still is) broken in 1048ab7c1434340e1793bc5765486f0a6d5c13a3.

TeslaRus:
thanks for info... next I will fix - alt / normal room collision filter (disable / enable function too). 
last commit - only tweens generation fix (waterfall in TR1 map 3)

vvs-:
Seems mature to me. All recent issues that previously appeared in this PR were fixed. #343, #386 and altroom4 are the only problems that were not addressed yet. But maybe that do not belong to this PR after all, I'm not sure.

TeslaRus:
Thanks for feedback - it helps to fix serious bugs in early stage

vvs-:
"altroom4 test can not been passed, because at same time only one room (in alt room sequence) is available for entities / rendering"

Then it should be changed, because in the Core engine this level just works as expected. So, obviously there is no such limitation. Camera there sees only one room at a time, depending on its direction relative to a portal.

PR #389, opened 2016-10-11
Base path

TeslaRus:
added command options to choose game data and resource path, to choose config and autoexec files;


TeslaRus:
TODO: add base_path to font loading script!

src/gameflow.h
```diff
@@ -43,7 +43,7 @@ class CGameflow
 
     uint8_t             getCurrentGameID();
     uint8_t             getCurrentLevelID();
-    const char*         getCurrentLevelName();
+    const char*         getCurrentLevelName(); ///@TODO: delete unused function!
```
Gh0stBlade: I don't recommend deleting this, getCurrentLevelName returns the current level's name. Will be useful for the far future when level end statistics etc will be shown.

TeslaRus: if it returns filename level name, it is duplication (maybe it is better to delete Engine_... version of that function); if it returns name like "Natla's mines" - it should be saved.

PR #391, opened 2016-10-21
zero normals

XProger:


Lwmte:
Interesting... Do you have any idea why it happens? I. e. there was special condition in, maybe, Core Design's AnimEdit utility to generate zero normals for certain vertices?

XProger:
sorry, my pull request is wrong and produce wrong result.
in original we have zero normals for many objects like movable blocks and I have no idea how original engine solve it (or using it for what)
image

vvs-:
Ok. But why not close it yourselves?

Issue #223, opened 2015-07-25
One way passable geometry


vvs-:
In *Temple of the Cat* geometry should be passable from room 81 sector 3,2 to room 69 sector 1,4, but not vice versa.

Lwmte:
Impossible to fix with current physics model.

Lwmte:
The whole problem here is current collisional approach is "sewing" together all the rooms, leaving passages where the doors should be. While I and TeslaRus initially thought it's going to work, eventually it didn't. There are few examples for this, particularly this ticket and #11.

What may work is emulating original collisional approach, when each entity bear particular room ID, and only this room ID is processed for collision. When room is switched via portal sector, collision filter is re-applied and entity uses new room ID mesh for collisions.

The problem here, is we potentially have *lots* of entities in a level, and each entity must process its own collisional routines. Not all of them, but at least dynamic ones - boulders, ragdolls, activated Damocles swords, etc. Also, entities may not overlap when placed in different overlapped rooms. So we must think of a way to process each entity's collision on an individual basis, regarding its current room ID.

We may re-use current collisional room mesh generator for actual heightmap calculation, but that part of code which generates room borders and doors has to go.

T4Larson:
Overlapping rooms seem to be pretty frequent throughout the levels, I've recently been flipped into the wrong room in hall.tr4 some times...
I remember having read that CrystalSpace had similar problems using Bullet. Since it is a portal-engine, it would also allow overlapping rooms - it seems they are going to tackle this with an extra plugin: See http://www.crystalspace3d.org/docs/online/manual/Physics-Bullet.html :
>At least one dynamic system should therefore be created. Usually, one would need one dynamic system per Crystal Space sector, the problem being to switch the dynamic systems when an object crosses a portal. This is not yet made automatically by the ‘Bullet’ plugin, this is however a feature of the WIP ‘physics2’ plugin. 

IMO to get this right in OpenTomb, it is necessary to track each entities current room number (almost as the old engine does it), and probably for every raycast, to ensure the collision is filtered for the correct set of neighbouring rooms...

Lwmte:
Hmm, so unfortunately it's Bullet who limits us here... But surely Erwin Coumans didn't take ancient sector-based pseudo-portalized engine into account when he programmed his physics! :laughing: 

[...]

PR #394, opened 2016-11-06
added dynamic altroom portals tweens generation;

TeslaRus:
now in every success  rooms flipping altroom portal tweens are rebuild-ed;
no more hacky logic: correct in all cases;

### DIFF
--- a/src/room.h
+++ b/src/room.h
@@ -214,7 +214,8 @@ typedef struct room_content_s
 
     float                       ambient_lighting[3];
     struct base_mesh_s         *mesh;                                           // room's base mesh
-    struct physics_object_s    *physics_body;
+    struct physics_object_s    *physics_body;                                   // static physics data
+    struct physics_object_s    *physics_body_dynamic;                           // static physics data
 }room_content_t, *room_content_p;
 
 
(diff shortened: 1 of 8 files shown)


src/room.h
```diff
@@ -214,7 +214,8 @@ typedef struct room_content_s
 
     float                       ambient_lighting[3];
     struct base_mesh_s         *mesh;                                           // room's base mesh
-    struct physics_object_s    *physics_body;
+    struct physics_object_s    *physics_body;                                   // static physics data
+    struct physics_object_s    *physics_body_dynamic;                           // static physics data
```
TeslaRus: fix comment, may be rename field (not good name)

vvs-:
It fixes _Angkor Wat_. I have found no regressions. But #223 is still valid of course.

TeslaRus:
#223 will be fixed / trying to fix in other branch, unfortunately I have no ideas how to fix it now...
that branch is only dynamic walls collision generation for rooms flipping (serious deal);

vvs-:
> unfortunately I have no ideas how to fix it now...
I'm not sure, but I think that once I've seen a bug where Lara was able to pass a collisional portal in one direction, but not in its opposite. I think this was some time during altroom_refactoring in _Tomb of Qualopec_. If I get more time, I'll try to bisect it.

vvs-:
Yes, I was right. This happened in 19d6085899e8c5d723eb352be7d4bb4031f974e7. But it was a special case when Lara climbed up from water. Still it could be an inspiration for a solution.

PR #399, opened 2016-11-16
decomposed entity scripts;

TeslaRus:
entity_functions.lua is too big for convenience working: it is decomposed to

- entity_functions.lua   (calls all scripts below, not sorted properly)
- entity_functions_common.lua   (doors, switches, keyholes, boulders)
- entity_functions_traps.lua   (blades, spikes)
- entity_functions_unique.lua   (unique single objects)

+ fixed tall block script;


vvs-:
`script-backporting` is falling behind more and more. I'm wandering, could it be merged at least partially? I mean piece by piece?

Lwmte:
I suggest that at least feature-critical code must be backported. I guess I'm too desperate on fixing timing and TRIGGERING_NOT_READY stuff.
Do I need to make pull request to make it happening?

TeslaRus:
I suggest to do backporting by LITTLE parts, to avoid big crashes, and when it become stable - use bigger backporting requests. +I fix many time stuff in entity scripts now.

TeslaRus:
@Lwmte : I create diff and study backporting branch. Really - it MUST be decomposed to 3 different requests minimum. It changes not only scripts, but some core functions (restores relative GUI size that should not be used in low level rendering! in addition separate module it may be correct, but not in core + GUI must be decomposed too and be placed to the gui folder - I will do it later). 
Current door script (master) realization is savable - no tasks and local lua vars;
Enemies scripts must be in separate lua file - easy to merge and develop;

Level scripts reorganization loading - one single request - let us begin from this one - middle size, single task - easy to verify;
Scripts backporting: first time by modules + cpp code changes should be in a different request (if it is possible of course) 

sorry, bu this request as a single shot is not pullable - only in decomposed state - we have no much time for reading and fixing request and next debugging merge results all time.

PR #401, opened 2016-11-22
Physics refactoring

TeslaRus:
C++ source goes after LUA scripts...


TeslaRus:
TODO: add TRIGGERS collision group; GHOST group ONLY for ghosts.

PR #403, opened 2016-11-27
Critical update: TR1-2 fixed cameras info

Lwmte:
Info verified with TR1-TR2 disassembly.


TeslaRus:
lgtm
I some info about cam sync playing in reverse direction? I.E. TR4 map4 rotated room

PR #414, opened 2016-12-31
Started translating notification strings in French, corrected some ot…

zdimension:
…hers

### DIFF
--- a/scripts/strings/french/generic.lua
+++ b/scripts/strings/french/generic.lua
@@ -1,4 +1,4 @@
--- OPENTOMB GENERIC STRINGS - ENGLISH LANGUAGE
+-- OPENTOMB GENERIC STRINGS - FRENCH LANGUAGE
 -- by Lwmte, Jan 2015
 -- French Translation by: Joey79100 and zdimension.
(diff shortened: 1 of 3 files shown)


scripts/strings/french/generic.lua
```diff
@@ -1,7 +1,7 @@
 -- OPENTOMB GENERIC STRINGS - ENGLISH LANGUAGE
```
TeslaRus: may be "French" in comment? ;-)

scripts/strings/french/sys_notify.lua
```diff
@@ -0,0 +1,67 @@
+-- OPENTOMB SYSTEM CONSOLE NOTIFICATIONS
```
TeslaRus: Sys notyfy not uses LUA strings: it is hardcoded... LUA translation ONLY for game and UI... system is only build in (English is hardcoded for sysstrings: reason: if something goes wrong in LUA then sys warning strings will be loosed + it is little bit slower)

scripts/strings/french/generic.lua
zdimension: Thanks! I didn't notice that. Fixed it in 26f6b76.

PR #415, opened 2017-01-03
Set anim control and collision fixing

TeslaRus:
- added auto position correction for Entity_SetAnimation function: for hardcoded anim state controlling without hardcoded magic offsets;
 - fixed Entity_GetPenetrationFixVector algorithm;
 - enabled anti-tunneling feature for Entity_GetPenetrationFixVector;

### DIFF
--- a/src/anim_state_control.cpp
+++ b/src/anim_state_control.cpp
@@ -829,7 +829,7 @@ int State_Control_Lara(struct entity_s *ent, struct ss_animation_s *ss_anim)
             else if((cmd->move[0] == 1) && (cmd->crouch == 0) && (next_fc.floor_hit.normale[2] >= ent->character->critical_slant_z_component) && (i == CHARACTER_STEP_UP_BIG))
             {
                 ent->dir_flag = ENT_STAY;
-                i = Anim_GetAnimDispatchCase(ent->bf, 2);                     // MOST CORRECT STATECHANGE!!!
+                i = Anim_GetAnimDispatchCase(ent->bf, 2);                       // MOST CORRECT STATECHANGE!!!
                 if(i == 0)
                 {
                     Entity_SetAnimation(ent, ANIM_TYPE_BASE, TR_ANIMATION_LARA_RUN_UP_STEP_RIGHT, 0);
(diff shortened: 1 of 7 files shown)


src/anim_state_control.cpp
```diff
@@ -2355,10 +2361,13 @@ int State_Control_Lara(struct entity_s *ent, struct ss_animation_s *ss_anim)
              * FREE FALL TO UNDERWATER CASES
              */
         case TR_STATE_LARA_UNDERWATER_DIVING:
-            ent->angles[1] = -45.0;
-            cmd->rot[1] = 0.0;
-            Entity_UpdateTransform(ent);
-            ss_anim->onEndFrame = ent_correct_diving_angle;
+            if(ent->move_type != MOVE_FREE_FALLING)
```
TeslaRus: maybe that condition or whole section is unnecessary...

vvs-:
Many entities has incorrect positions, e.g. doors and switches.

TeslaRus:
wow! fast feedback! hope it https://github.com/opentomb/OpenTomb/commit/3b21f78997795578196383fcd60493d145de1856 helps

vvs-:
So far so good. But the position at the end of ladder climbing is still incorrect. Also, there are problems with wading on slopes.

TeslaRus:
Ladder fixing is another task (walls climbing only); that task is collision fixes fix and smooth positioning (hang to free fall: Lara falls down without up moving ~100 game units)
"Also, there are problems with wading on slopes." it is old bug meaning?

vvs-:
> "Also, there are problems with wading on slopes." it is old bug meaning?

It's old, but this PR introduced new manifestations which I didn't see before.

Also, this PR breaks underwater propeller in TR2 _Offshore Rig_.

TeslaRus:
I will test propellers... interesting... maybe reason of new bugs is that collision fixes works more correct so old bugs become visible (really before that PR was big mistake that prevents to "fix" out of room geometry).

vvs-:
Sorry, it was my confusion. Actually, that propeller bug isn't new to this PR. This is happening because it's animated when flags are 0 instead of 31.

PR #416, opened 2017-01-07
Update codeblocks project to support OpenAL soft extensions again

Lwmte:
Also remove super-old integrated AL sources from project and fix config path

NOTE: I didn't fix OpenAL paths, cause seems it works as is for everyone except me

### DIFF
--- a/src/audio.cpp
+++ b/src/audio.cpp
@@ -4,7 +4,7 @@
 
 #include <math.h>
 
-#include "config-default/config-opentomb.h"
+#include "config-opentomb.h"
 
 extern "C" {
 #include <al.h>
--- a/src/audio.cpp
+++ b/src/audio.cpp
@@ -11,7 +11,6 @@ extern "C" {
 #include <alc.h>
 #ifdef HAVE_ALEXT_H
 #include <alext.h>
-#include "config-default/config-opentomb.h"
 #endif
 #ifdef HAVE_EFX_H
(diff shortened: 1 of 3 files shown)


vvs-:
Seems that it doesn't compile.

src/audio.cpp
```diff
@@ -4,13 +4,14 @@
 
 #include <math.h>
 
-#include "config-opentomb.h"
+#include "config-default/config-opentomb.h"
```
TeslaRus: may be it is a reason of Travis failed...

vvs-: Yes, that's wrong. There are platform dependent includes which won't work on another platform.

TeslaRus:
here (https://github.com/TeslaRus/OpenTomb) AL reverb is enabled (C::B project is actual) and sources may be updated from main repo just by copy - replace all - all works

src/audio.cpp
```diff
@@ -11,6 +11,7 @@ extern "C" {
 #include <alc.h>
 #ifdef HAVE_ALEXT_H
 #include <alext.h>
+#define AL_ALEXT_PROTOTYPES
```
TeslaRus: should be in **config-opentomb.h**
after changes may be merged

Lwmte: done ;)

TeslaRus:
lgtm

PR #423, opened 2017-02-14
preparing for enemies state control implementation;

TeslaRus:


vvs-:
`mlook()` doesn't work.

TeslaRus:
I disable no mlook option unlil no hacky solution will be founded (this time it uses Lara's states and anims checkings - nasty architecture, that I made some years ago...). I am trying to fully encapsulate animations and states (characters) inside state control module (onwater to dive disabled too by same reason)...
Branch will be merged after auto dive restoring (and may be after mlook restoring)

TeslaRus:
mlook() is restored now; auto dive (uw current) does not work;

vvs-:
UW current was already broken. This is even scarcely documented for TR2 progress.

src/state_control/state_control_Lara.cpp
```diff
@@ -196,7 +196,31 @@ void ent_crawl_to_climb(entity_p ent, ss_animation_p ss_anim)
 }
 
 
-int State_Control_Lara(struct entity_s *ent, struct ss_animation_s *ss_anim)
+void StateControl_LaraSetDefaultAnim(struct entity_s *ent, int anim_type, int move_type)
```
stohrendorf: Shouldn't this be better named `...ApplyIdleAnim` or something similar?

src/state_control/state_control_Lara.cpp
```diff
@@ -196,7 +196,31 @@ void ent_crawl_to_climb(entity_p ent, ss_animation_p ss_anim)
...
+        case MOVE_UNDERWATER:
+            Entity_SetAnimation(ent, anim_type, TR_ANIMATION_LARA_UNDERWATER_IDLE, 0, NULL);
+            break;
+
+        case MOVE_ON_WATER:
+            Entity_SetAnimation(ent, anim_type, TR_ANIMATION_LARA_ONWATER_IDLE, 0, NULL);
+            break;
+
+        case MOVE_ON_FLOOR:
+        default:
```
stohrendorf: This silently swallows wrong values of `move_type`.

src/character_controller.cpp
```diff
@@ -45,12 +44,15 @@ void Character_Create(struct entity_s *ent)
...
-        ret->resp.horizontal_collide = 0x00;
-        ret->resp.kill = 0x00;
-        ret->resp.burn = 0x00;
-        ret->resp.slide = 0x00;
-        ret->resp.step_z = 0x00;
+        ret->state.vertical_collide = 0x00;
+        ret->state.horizontal_collide = 0x00;
+        ret->state.slide = 0x00;
+        ret->state.step_z = 0x00;
+        ret->state.kill = 0x00;
```
stohrendorf: Is `kill` a request or a state? If it's a state, I'd rather call it `dying` or `dead`.

src/character_controller.cpp
```diff
@@ -1249,7 +1251,7 @@ int Character_FreeFalling(struct entity_s *ent)
         {
             pos[2] = ent->character->height_info.ceiling_hit.point[2] - ent->bf->bb_max[2];
             ent->speed[2] = 0.0f;
-            ent->character->resp.vertical_collide |= 0x02;
+            ent->character->state.vertical_collide |= 0x02;
```
stohrendorf: Maybe replace the OR'd values with some meaningful constant names.

src/character_controller.h
```diff
@@ -237,15 +237,18 @@ typedef struct character_command_s
...
+typedef struct character_state_s
 {
     int8_t      vertical_collide;
     int8_t      horizontal_collide;
-    uint16_t    kill : 1;
-    uint16_t    burn : 1;
     uint16_t    slide : 2;
     uint16_t    step_z : 2;
-}character_response_t, *character_response_p;
+    uint16_t    kill : 1;
```
stohrendorf: Most of these are used as booleans, which is not clear from the code; others are used as enums, and others are used as bitfields. If it's not possible (or too much work) to use proper data types, please add at least a few comments here what each member is supposed to contain and which effects it should have.

src/character_controller.cpp
TeslaRus: kill -> dead, better as state than request; character->kill() isa request, but it is not a class...

src/character_controller.h
TeslaRus: yep, first comments, maybe later it will be more clear what to do with architecture

src/state_control/state_control_Lara.cpp
TeslaRus: IdleAnim more suits here

vvs-:
Ok. She now dives underwater, but that's all. The implementation is even more broken than it was before. Now Lara jumps out of water sometimes.

Also, look at e.g. _Living Quarters_ where she's unable to proceed from the starting area. Though, I guess that there is just not enough information about how it worked in the original engine in the first place.

vvs-:
Much better. But now it badly interacts with the camera. Look at waterfall in _Lost Valley_, there is a constant jitter of POV between two cameras when trigger is activated.

vvs-:
Mostly works ok, but sometimes still has camera problems in rare cases.

PR #427, opened 2017-03-17
If the config.lua file is not found then there is

devnexen:
a logging attempt donw but it leads to segfaulting
since some allocations done in Engine_Init_Pre
are needed for it, thus swapping their places here.

### DIFF
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -177,10 +177,10 @@ void Engine_Start(int argc, char **argv)
         }
     }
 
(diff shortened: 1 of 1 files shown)


devnexen:
I found it while I usually create a separated folder with cmake based projects and forgot to take necessary files like config.lua etc ... In fact the segfault happens because Con_Init is not called yet.

TeslaRus:
lgtm, thanks for fix;
I am planning to make Con module init before config loading (by making lines size tunable in real time after initialization);

PR #428, opened 2017-04-02
Physics refactoring

TeslaRus:
- physics modules now are in **physics** dir;
- ragdoll and hair moved to their own modules (inside **physics** dir);
- stabilized collision resolving and deleted no_fix_z collision resolving flag;
- ragdoll and hair LUA parsers now are more abstract (less code dependencies);


vvs-:
Something is broken. Either a jumping trajectory become lower or Lara has now legs collision where it wasn't before. She can't make a running jump to a timed double door in _Caves_.

TeslaRus:
yep, it is no_fix_z disabling... should be solved by the other way

vvs-:
There is another problem. Lara could disappear in noclip mode when out of any room.

TeslaRus:
> Lara could disappear in noclip mode when out of any room.

 - it is normal... Lara become usual entity ~4 month ago...

PR #431, opened 2017-04-06
Proposal, MSVC 2017 compiler compatibility patches.

Gh0stBlade:
Open to suggestions on how to improve this too.

Maybe a custom align macro to prevent duplicated lines.


vvs-:
Travis suggests to add `#include` for memset in frustum.cpp.

stohrendorf:
1) It adds a lot of #ifdefs for alignments, maybe make a global typedef for them.
2) also, there are a few useless #ifdefs with VLAs, personally I'd prefer to only have the cross-compiler compatible branch.

TeslaRus:
ok, I will use some fixes: PR is an instruction, not for merge.
thanks: here is some interesting fixes

Gh0stBlade:
@vvs- Yep, I did at the missing header but Travis still complains.
@stohrendorf You are very right about the annoying ifdefs, defining a macro or typedef is definitely the better solution.
@TeslaRus 
That's fine 👍 Can you close the PR once you take the patches you want. The only change of interest I can think of is the gameflow stuff. Everything else is identical, just changed slightly to compile under MSVC++ compiler.
===

I will continue to develop features in this branch from now on. I'm not a fan of using Mingw32/Codeblocks, VS is better for me and I'm sure it could attract more interest. 

Cheers.

src/render/frustum.cpp
```diff
@@ -1,5 +1,6 @@
 
 #include <stdio.h>
+#include <string>
```
vvs-: It should be a `cstring` in gcc.

src/core/system.c
```diff
@@ -105,20 +110,46 @@ void Sys_ResetTempMem()
...
 */
-float Sys_FloatTime (void)
+
+#if _MSC_VER
+typedef struct timeval {
+	long tv_sec;
+	long tv_usec;
+} timeval;
+
+int gettimeofday(struct timeval * tp, struct timezone * tzp)
```
TeslaRus: there are: 
 - SDL_GetTicks() // bad precision: ms - do not use;
 - SDL_GetPerformanceCounter() and SDL_GetPerformanceFrequency() // is SDL_GetPerformanceFrequency() slow, can I use it once, on app init?

stohrendorf: I _think_ SDL_GetPerformanceFrequency may change when the processor changes its frequency.

src/audio.cpp
```diff
@@ -1,6 +1,5 @@
-
-#include <SDL2/SDL.h>
-#include <SDL2/SDL_audio.h>
+#include <SDL2/SDL_platform.h>
+#include <SDL2/SDL_opengl.h>
```
TeslaRus: OpenGL in audio??? If GLfloat, it should be replaced by float

Gh0stBlade: ALfloat! pretty sure ALfloat/GLfloat are just typedefs for float anyway

src/world.cpp
```diff
@@ -1441,8 +1441,8 @@ void World_GenAnimTextures(class VT_Level *tr)
             seq->frame_lock        = false; // by default anim is playing
             seq->uvrotate          = false; // by default uvrotate
             seq->reverse_direction = false; // Needed for proper reverse-type start-up.
-            seq->frame_rate        = 0.05;  // Should be passed as 1 / FPS.
-            seq->frame_time        = 0.0;   // Reset frame time to initial state.
+            seq->frame_rate        = 0.05f;  // Should be passed as 1 / FPS.
```
TeslaRus: done all in world.cpp

src/gameflow.h
```diff
@@ -64,7 +64,7 @@ class CGameflow
     char                m_currentLevelName[LEVEL_NAME_MAX_LEN];
     char                m_currentLevelPath[MAX_ENGINE_PATH];
 
-    gameflow_action     m_actions[GF_MAX_ACTIONS];
```
TeslaRus: gameflow was rewritten... but it still not complete...

src/character_controller.cpp
```diff
@@ -227,13 +227,13 @@ void Character_UpdateCurrentSpeed(struct entity_s *ent, int zeroVz)
  */
 void Character_UpdateCurrentHeight(struct entity_s *ent)
 {
-    float from[3], to[3], base_z, *v;
+    float from[3], *v;
```
TeslaRus: done

src/character_controller.cpp
```diff
@@ -227,13 +227,13 @@ void Character_UpdateCurrentSpeed(struct entity_s *ent, int zeroVz)
...
     height_info_p hi = &ent->character->height_info;
 
     v = ent->bf->bone_tags[0].transform + 12;
     Mat4_vec3_mul_macro(from, ent->transform, v);
     from[2] += 128.0f;
-    base_z = from[2];
+
```
TeslaRus: done

src/character_controller.cpp
```diff
@@ -969,7 +969,7 @@ int Character_MoveOnFloor(struct entity_s *ent)
 
     norm_move_xy[0] = move[0];
     norm_move_xy[1] = move[1];
-    norm_move_xy_len = sqrt(move[0] * move[0] + move[1] * move[1]);
+    norm_move_xy_len = sqrtf(move[0] * move[0] + move[1] * move[1]);
```
TeslaRus: done

TeslaRus:
main fixes was added... other things should be done separately;

PR #444, opened 2017-08-10
Add option to invert mouse x/y axis

johnimorris:
Added options mouse_invert_x and mouse_invert_y that allow the x or y axis to be reversed for mouse camera control.

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -36,6 +36,8 @@ render =
 controls =
 {
     mouse_sensitivity = 25.0;
+    mouse_invert_x = 0;
+    mouse_invert_y = 0;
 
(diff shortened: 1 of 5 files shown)


TeslaRus:
Hallow! Thanks for idea =) 
I have found that it may be better to use mouse_sensitivity_x and mouse_sensitivity_y (no mouse_sensitivity) and negative sensitivity value to invert : no need additional conditions in  src/engine.cpp

TeslaRus:
feature added in https://github.com/opentomb/OpenTomb/commit/ebd74685258ff8a14b4781330167719c48bdbcea

PR #467, opened 2017-10-14
few code fixes

devnexen:
- possible buffer overflow fixes.
- proper type key for items_tree.


src/character_controller.cpp
```diff
@@ -492,6 +492,8 @@ void Character_UpdateCurrentHeight(struct entity_s *ent)
     v = ent->bf->bone_tags[0].transform + 12;
     Mat4_vec3_mul_macro(from, ent->transform, v);
     from[2] -= ent->speed[2] * engine_frame_time;
+    from[0] = ent->transform[12 + 0];
+    from[1] = ent->transform[12 + 1];
```
vvs-: Already fixed in 3ace169282b7fe4dd958559d9508121445fb10f0.

devnexen: True I think this might be wrong rebasing from me here

PR #469, opened 2017-10-15
Improved inclusion of OpenAL files

carlo-bramini:
I tried to compile OpenTomb with mingw-w64 under CYGWIN.
During this process I had got an error because it could not find OpenAL files.
This happened because in CYGWIN the path to the compiler does not match only to the compiler name, but it also includes an extra "/sys-root/mingw".
I did a fix into Toolchain-cross-mingw32-linux.cmake by declaring two different constants, one for the toolchain name prefix and one for the toolchain path prefix.
In this commit, I left previous situation and I kept commented the fix to CYGWIN. To get it working with this enviroment, the proper macro needs to be uncommented, in a similar manner it had to be done with the compiler name.
However, I would like to suggest to change the inclusion of openal files from:

#include <al.h>

to:

#include <AL/al.h>

by using the same convention tipically used with OpenGL, for example.
For this purpose I added src/audio/audio_al.h, that it is included in the various source files for simplify a bit. After that, the extra line in the configuration file for mingw cross compilation for adding a search path to the openal include files become useless, so it has been removed.

Sincerely.

PS: I do not own the games, so I tried to run the demos but I could not. Is it a known limitation/bug or I have to follow a particular procedure?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,7 +182,6 @@ if(MINGW AND CMAKE_CROSSCOMPILING)
     # Cross compiling on Linux with the MinGW toolchain.
     message(STATUS "MinGW Cross-Compilation")
 
-    set(OPENAL_INCLUDE_DIR "/usr/${COMPILER_PREFIX}/include/AL")
     set(OPENAL_LIBRARY OpenAL32)
 
     set(SDL2_INCLUDE_DIR "/usr/${COMPILER_PREFIX}/include/SDL2")
--- a/cmake/Toolchain-cross-mingw32-linux.cmake
+++ b/cmake/Toolchain-cross-mingw32-linux.cmake
@@ -5,16 +5,22 @@
 SET(CMAKE_SYSTEM_NAME Windows)
 
 # the classical mingw32 (http://www.mingw.org/)
-#SET(COMPILER_PREFIX "i586-mingw32msvc")
+#SET(TOOLCHAIN_PREFIX "i586-mingw32msvc")
 
 # 32 or 64 bits mingw-w64 (http://mingw-w64.sourceforge.net/)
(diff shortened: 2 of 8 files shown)


vvs-:
This is a quote from FindOpenAL.cmake:
```
# This makes the presumption that you are include al.h like
# #include "al.h"
# and not
# #include <AL/al.h>
# The reason for this is that the latter is not entirely portable.
# Windows/Creative Labs does not by default put their headers in AL/ and
# OS X uses the convention <OpenAL/al.h>.
```

TeslaRus:
Demos not implemented yet... TR1 cut scenes - implemented; RPL video playing implemented (but not implemented scripts to call it); Audio module will be refactored / updated (I am planning to hide AL part inside cpp, in the interface modules no one AL includes / types...). About CMake update - good, acceptable, but not audio module. I want to restore soft AL using (it works under MinGW, Linux and CygWin) + add one option in one place to switch internal AL and system AL.

carlo-bramini:
I see.
Do you prefer that I revert completely the changes on the inclusion of openal files or I could keep the src/audio/audio_al.h and just remove the "AL/" prefix from the file names?

Sincerely.

TeslaRus:
Hi! It is better to "revert completely the changes on the inclusion of openal files...", audio headers will be updated (include part rewritten)

TeslaRus:
TODO for myself (do not forget): move AL initialization outside engine.cpp to the audio. Clean includes from AL...

carlo-bramini:
This pending request has been superseded by #473, to make changes simpler.
Sincerely.

TeslaRus:
done in
Added Linux and Windows CI builds using GitHub Actions #601

Issue #378, opened 2016-08-28
Inverted Y axis while swimming


Joey79100:
Minor issue when playing with a gamepad and using analog sticks to move Lara: while swimming underwater, pushing the joystick up should make Lara go down, and pushing it down should make her go up: instead it is happening the opposite, while there's no problem with bound keys.

Pritchy96:
I think I'll take a look at this, if that's okay? Seems like a good entry point.

PR #471, opened 2017-10-16
Fix inverted swimming gamepad controls while underwater (#378)

Pritchy96:
A very easy, self explanatory fix for #378. 

I did a quick search and `ent->character->cmd.rot[1]`seems to be used only for movement underwater, so I can't foresee any issues caused by this change elsewhere.


TeslaRus:
I do not use joystick, so...

Issue #495, opened 2018-01-06
Translations for Portuguese (Portugal)


filfreire:
Do similar work to PR #414 but for Portuguese (Portugal) language.

/cc @TeslaRus @Gh0stBlade I'd like to submit a PR for this.

PR #497, opened 2018-01-06
Add translation Portuguese - Portugal #495

filfreire:
- plus a small fix for English generic translation

/cc @TeslaRus @Gh0stBlade 

related Issue #495


scripts/strings/portuguese-ptpt/sys_notify.lua
```diff
@@ -0,0 +1,67 @@
+-- OPENTOMB SYSTEM CONSOLE NOTIFICATIONS
```
TeslaRus: console output and system messages now is hardcoded and English only to avoid broken / empty debug output in case of script crash or other heavy bugs...

PR #521, opened 2018-02-16
Update Readme.md

agentzero1:
Filipino Translation


TeslaRus:
Please move translated readme.md to the Languages/es/readme.md
(is "es" correct name?)

agentzero1:
Filipino , so I use tl right ?

PR #522, opened 2018-02-16
Filipino translation of README.md

agentzero1:
Filipino translation


agentzero1:
I use tl because its Filipino am I right ?

PR #529, opened 2018-02-18
Filipino Translation of LICENSE

agentzero1:


vvs-:
No, no. This pull request is wrong either. Now you are replacing the original English license with your translation. You should keep both the original file and your translation in a separate folders.

Issue #534, opened 2018-02-25
Type mismatch in adpcm_data


carlo-bramini:
Hello,
I did a full c++ compilation, including the c sources and I noticed some things.
One of these is a type mismatch for ff_adpcm_afc_coeffs[][] into adpcm_data.
Here, it is declared as int16_t:

https://github.com/opentomb/OpenTomb/blob/1fe14f278b36f6bc3438453bfa4e9fd8ccddc1de/src/fmv/codecs/adpcm_data.h#L43

but here it is declared as uint16_t:

https://github.com/opentomb/OpenTomb/blob/1fe14f278b36f6bc3438453bfa4e9fd8ccddc1de/src/fmv/codecs/adpcm_data.c#L109

Perhaps it would be worth to fix this, by declaring the values in the table as signed integers, like:

```
{ 0, 2048, 0, 1024, 4096, 3584, 3072, 4608, 4200, 4800, 5120, 2048, 1024, -1024, -1024, -2048 }
{ 0, 0, 2048, 1024, -2048, -1536, -1024, -2560, -2248, -2300, -3072, -2048, -1024, 1024, 0, 0 }
```
I could provide this patch if you think that it would be useful.

TeslaRus:
merged

PR #535, opened 2018-03-03
Type mismatch during full C++ build (Windows, MSVC).

carlo-bramini:
This patch closes issue #534.
Sincerely.

### DIFF
--- a/src/fmv/codecs/adpcm.c
+++ b/src/fmv/codecs/adpcm.c
@@ -1253,7 +1253,7 @@ static int32_t adpcm_decode_frame(struct tiny_codec_s *avctx, struct AVPacket *a
             int coeff1, coeff2;
             int shift;
             unsigned int channel;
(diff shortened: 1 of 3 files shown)


TeslaRus:
Hi! Interesting type mismatch, but it is copy of original ffmpeg library. https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/adpcm_data.c and https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/adpcm_data.h
I have to check first will it break original algorithm.
**extern "C"** is not necessary: header is used internally only in .c file

carlo-bramini:
Hello. Actually, I had to introduce extern "C" because I have built OpenTomb by forcing C++ to all source. Otherwise, the various tables were not visibile outside. Perhaps it would be worth to signal that issue directly to FFmpeg team, so that you could just sync your sources with their master if they will accept the change.

PR #541, opened 2018-03-18
Add selection of system LUA and BULLET libs

carlo-bramini:
This patch includes support for selecting system provided LUA 5.3 (and newer) and BULLET, in a similar manner it had been done for FreeType.

* Added two new options, FORCE_SYSTEM_LUA and FORCE_SYSTEM_BULLET. By default, they are OFF so the internal copies or the libraries will be used without parameters. LUA is accepted only if version 5.3 or newer is found.

* Added a new file,  src/core/lua.h, used for including LUA stuff.

* Inside src/core/lua.h, include directive has been changed from:
`#include <lua.h>`
to:
` #include "lua.h"`
as it has been written in the CMAKE documentation:
https://cmake.org/cmake/help/v3.0/module/FindLua.html

Tested with MINGW-W64 cross compiler: LUA 5.2 is rejected, BULLET is not found.
Tested with MSVC Community 2017: LUA 5.3 and BULLET are found and accepted.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,8 @@ project(OpenTomb)
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 
 option(FORCE_SYSTEM_FREETYPE "Use system-provided FreeType instead of internal library." OFF)
+option(FORCE_SYSTEM_LUA      "Use system-provided LUA 5.3+ instead of internal library." OFF)
+option(FORCE_SYSTEM_BULLET   "Use system-provided BULLET instead of internal library."   OFF)
 
 # Detect system FreeType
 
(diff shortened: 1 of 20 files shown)


TeslaRus:
thanks, good idea... I will try to review the changes.
minimum LUA version is 5.3

carlo-bramini:
I have set a variable called LUA_MIN_VERSION for checking the minimum version of the LUA package.
If, in the future, there will be the need to use a more recent version than 5.3, you can just change that variable.

RICCIARDI-Adrien:
Hi,
I tried on Linux your CMakeLists.txt modifications for using system-provided Lua, but I did not modify `lua.h` header inclusion. Everything worked fine on Xubuntu 17.10. Is it also working without `lua.h` modifications on Windows ?
I'm thinking that CMake documentation is telling that specifying "lua" directory when including `lua/lua.h` is not portable, but for me using `#include "lua.h"` or `#include <lua.h>` here does not change anything (both system-provided and OpenTomb-provided lua are working on my Linux distro).

carlo-bramini:
I was forced to change #include <lua.h> to #include "lua.h", othewise I got an error when compiling with VisualStudio 2017.

RICCIARDI-Adrien:
Hi @carlo-bramini, maybe you can do another pull request containing bullet-related modifications only, so they get included (they were working fine on Linux). Lua stuff could be done in a second time.

CMakeLists.txt
```diff
@@ -10,6 +10,8 @@ project(OpenTomb)
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 
 option(FORCE_SYSTEM_FREETYPE "Use system-provided FreeType instead of internal library." OFF)
+option(FORCE_SYSTEM_LUA      "Use system-provided LUA 5.3+ instead of internal library." OFF)
```
Iniquitatis, line 13: Not LUA, but Lua ([reference](https://www.lua.org/)).

Issue #546, opened 2018-04-29
Embed TR1-TR2 loading screens in OpenTomb repo


RICCIARDI-Adrien:
Here is a proposition to ease TR1 and TR2 games setup.
Loading screens provided in README.md links come in uppercase JPEG files, whereas the engine is expecting lowercase PIX or PNG. On all platforms you have to convert JPEGs to PNG, and on Linux you must rename them in lowercase.
As TR1 and TR2 loading screens does not exist in commercial version, I think there is no problem providing them with OpenTomb.

PR #549, opened 2018-05-22
Added TR1 loading screens.

RICCIARDI-Adrien:
Here are TR1 loading screens as mentioned in issue #546. TR2 screens will follow if this pull request is accepted.


TeslaRus:
Hallow! It is better to use different place for resources like load screens, maps a.t.c.... +TR1 load screens may be copied from original CD (pcx format is supported).
It is may be better to provide solution for resource copy / rename from original CD or gog / steam installation.

RICCIARDI-Adrien:
Ok, no problem.
I close this pull request and the related issue.

Issue #561, opened 2018-10-13
Ubuntu: System error: Could not create SDL window


neo-expert:
OpenTomb compiles sucessfully, but after running "./OpenTomb" it says:
System error: Could not create SDL window. On Ubuntu and on Elementary OS.

what can I do?

Gh0stBlade:
Perhaps there's a driver issue on your system? Any further details about system specifications, driver versions and OS version should help someone else paint a bigger picture of what's going on here.

kgorecki:
Most probably you don't have GLX visual set. Please check this solution: https://stackoverflow.com/questions/41338549/sdl2-cant-create-window-since-it-couldnt-find-matching-glx-visual
I think that some better error message would be good here, "Could not create SDL window" doesn't give you much info. What about using SDL_GetError() instead?

RICCIARDI-Adrien:
Hi,
Testing on an old Pentium 4 with a GeForce 6200 running Debian Testing I think I got the same error :
```
System error: Could not create SDL window: Couldn't find matching GLX visual
```
Error might be related to antialiasing, because the error changes when I disable this feature in config file.
With `antialias = 0;` in config file, I get :
```
Error opening shaders/static_mesh.vsh
Abandon
```
I will try to investigate more but I know nothing about shaders...

kgorecki:
Did you check default configuration with applied instructions from link above?

neo-expert:
Yeah with `antialias = 0;` it works now. Thank you very much.

But I have a bit more modern hardware:
Radeon HD 8970M with Intel(R) Core(TM) i7-4800MQ CPU @ 2.70GHz

kgorecki:
It really starts without antialiassing. However I'd recommend to configure your OS properly. Typing 2 commands (3 if you don't have glxinfo by default) doesn't hurt. If you really deny to check the link, here's the instruction:
Ensure that GLX is installed correctly by running glxinfo. At the bottom, you'll find the list of supported visuals. Here's mine:

1 GLX Visuals
    visual  x   bf lv rg d st  colorbuffer  sr ax dp st accumbuffer  ms  cav
  id dep cl sp  sz l  ci b ro  r  g  b  a F gb bf th cl  r  g  b  a ns b eat
----------------------------------------------------------------------------
0x022 24 tc  0  24  0 r  y .   8  8  8  0 .  .  0 16  0  0  0  0  0  0 0 None

Try running the following before running the SDL2 program:

[...]

PR #563, opened 2018-10-24
A bit more verbose SDL errors

kgorecki:
Like it is mentioned in an issue #561 it's quite hard to get anything about an issue with SDL errors. I've added a bit more info in Sys_Error calls in the Engine.

### DIFF
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -260,7 +260,6 @@ const char *Engine_GetBasePath()
 const char *Engine_GetSDLError(const char* message)
 {
     const char *sdlMessage = SDL_GetError();
-//    char message[] = "Could not create SDL window: ";
     char *errorMessage = new char[strlen(message) + strlen(sdlMessage)];
     strcpy(errorMessage, message);
     strcat(errorMessage, sdlMessage);
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -257,17 +257,6 @@ const char *Engine_GetBasePath()
 }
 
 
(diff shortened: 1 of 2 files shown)


src/engine.cpp
```diff
@@ -257,6 +257,17 @@ const char *Engine_GetBasePath()
 }
 
 
+const char *Engine_GetSDLError(const char* message)
+{
+    const char *sdlMessage = SDL_GetError();
+    char *errorMessage = new char[strlen(message) + strlen(sdlMessage)];
+    strcpy(errorMessage, message);
+    strcat(errorMessage, sdlMessage);
```
stohrendorf: This will write the NUL character out-of-bounds, because the buffer is 1 character too short.

Apart from that, I agree with @TeslaRus.

TeslaRus: nice!

src/engine.cpp
```diff
@@ -257,6 +257,17 @@ const char *Engine_GetBasePath()
 }
 
 
+const char *Engine_GetSDLError(const char* message)
+{
+    const char *sdlMessage = SDL_GetError();
+    char *errorMessage = new char[strlen(message) + strlen(sdlMessage)];
```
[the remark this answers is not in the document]
kgorecki: I'm embarrassed... Moved to format string.

src/engine.cpp
kgorecki: It printed correctly even if you're right that it's 1 char short. Done it without reinventing the wheel though.

PR #566, opened 2018-11-05
ENGINE.md: Fixed typos

fraang:
`tate*` to `state*`


TokyoSU:
Hello @fraang,

@TeslaRus can be fix it, so you can just submit a issue, if is just to rewrite 3 line you dont need to write a pull request for this, but gg to find this :)

PR #568, opened 2018-11-22
fix useless missing ";" in lua scripts

TokyoSU:
@TeslaRus you have see that lua scripts folder have missing ";"
and i fixed it for you and added or rewrite some print(""); comment to discovers the files as being loaded !
(no level script is included because too mush :x )


TeslaRus:
lgtm
wow! big good work, thanks!
review had been done. may be merged now (will be new commits soon?).
is old "entity_model_ID_override.lua" renamed?

TokyoSU:
i want to rewrite all level name (files), like WALL to LEVEL1 but when i change the pull request to add the new files, the old files are not deleted :x,
can i change name, i will change gameflow scripts too for sure :) ?

TokyoSU:
You can merge it now, if you accept to add an option to load script with gameflow i change it for you, for now script is clean and level have the original name (in print()) (not for TR2)

PR #586, opened 2019-03-10
Fixed '-Wformat-truncation' gcc warning.

RICCIARDI-Adrien:


src/engine.cpp
```diff
@@ -1382,8 +1382,8 @@ extern "C" int Engine_ExecCmd(char *ch)
             }
             else
             {
-                char buf[1024];
-                snprintf(buf, 1024, "Command \"%s\" not found", token);
+                char buf[1088]; // Give some more room to store the printf() format string plus the token argument (this avoids triggering -Wformat-truncation gcc warning)
```
TeslaRus: ok, but what is 1088? is it GCC specific constant, or magic?

RICCIARDI-Adrien: The printf() format string is about 21-character long, so I reserved 64 bytes in case someone changes it. 1088 is 1024 + 64 (1024 is `token` maximum size).

TeslaRus: ok, thanks )

PR #591, opened 2019-05-14
Added weapons short keys support.

RICCIARDI-Adrien:


TeslaRus:
Thanks for PR, I will see it little bit later...
P.S. first look - all is ok.

src/weapons.cpp
```diff
@@ -1,5 +1,6 @@
 
 #include "audio/audio.h"
+#include "controls.h"
```
TeslaRus: hmmm ok, I know how to do it without this include )

RICCIARDI-Adrien: Hi !
How could this have been done ?
I'm interested in OpenTomb internal architecture.

TeslaRus: done in my commit - just substract weapon code outside function, that uses it  ))

PR #592, opened 2019-05-18
Update Windows Compiler Compatibility

TokyoSU:
the attributes(packed, aligned) not exist on windows compiler (stdc++17),
because, if bullet not crash, windows compiler will align the variable correctly.
i added some fix like config-opentomb for no cmake user, lua path "AL/".
and added new cpp: "timer.cpp" that fix clock_gettime() and gettimeofday (and fix CLOCK_REALTIME too),
and fixed "char image_name_buf[]" not compatible with c++17 (maybe need check about it).

### DIFF
--- a/src/gui/gui.cpp
+++ b/src/gui/gui.cpp
@@ -555,7 +555,7 @@ bool Gui_LoadScreenAssignPic(const char* pic_name)
     size_t pic_len = strlen(pic_name);
     size_t base_len = strlen(Engine_GetBasePath());
     size_t buf_len = pic_len + base_len + 5;
-    char image_name_buf[buf_len];
+    char *image_name_buf = new char[buf_len];
     int image_format = 0;
 
     strncpy(image_name_buf, Engine_GetBasePath(), buf_len);
--- a/src/render/camera.h
+++ b/src/render/camera.h
@@ -66,9 +66,15 @@ typedef struct camera_state_s
 
 typedef struct camera_s
 {
+#ifdef __GNUC__
     GLfloat                     gl_view_mat[16] __attribute__((packed, aligned(16)));
     GLfloat                     gl_proj_mat[16] __attribute__((packed, aligned(16)));
     GLfloat                     gl_view_proj_mat[16] __attribute__((packed, aligned(16)));
+#else
+    GLfloat                     gl_view_mat[16];
+    GLfloat                     gl_proj_mat[16];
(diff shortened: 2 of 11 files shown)


Gh0stBlade:
Pretty sure my solution was more elegant, did TeslaRus not take all of my patches for MSVC?

vvs-:
For reference it was in #431. What exactly was merged is hard to say, but it looks like it was done on 7-8 April, 2017, so those commits might be worth looking at.

TokyoSU:
it not exist in the src of opentomb, sadly @TeslaRus not merge all the change in #431

src/gui/gui.cpp
```diff
@@ -555,7 +555,7 @@ bool Gui_LoadScreenAssignPic(const char* pic_name)
     size_t pic_len = strlen(pic_name);
     size_t base_len = strlen(Engine_GetBasePath());
     size_t buf_len = pic_len + base_len + 5;
-    char image_name_buf[buf_len];
+    char *image_name_buf = new char[buf_len];
```
TeslaRus: 1) Where is delete[]?
2) char *image_name_buf = (char*)Sys_GetTempMem(buf_len);

src/render/camera.h
```diff
@@ -66,9 +66,15 @@ typedef struct camera_state_s
 
 typedef struct camera_s
 {
+#ifdef __GNUC__
```
TeslaRus: so much copy-paste code (original OT source problem)...
hmmm OK, I will fix `__attribute__((packed, aligned(16)))` soon
timer is more complex problem (P.S. VS has some posix like time functions; P.P.S. `QueryPerformanceFrequency` is inside SDL)

src/gui/gui.cpp
TokyoSU: hmm can be that the problem to some small freeze in game ? :x

src/render/camera.h
TokyoSU: you can close it anytime, it's not really needed, more like references for windows error at compile :)

TeslaRus:
partially done in 6311fa374504efbb5494b9485990fbb5154da181
new timer - nope...
using threads in codec module - stupid idea (my idea)... so timer / threading in codecs will be deleted.

TeslaRus:
timer will be fixed by other way.
thanks for PR

Issue #538, opened 2018-03-13
Crash on Introduction Level


Sur3:
Hi I tried playing OpenTomb from master yesterday, setgamef(1,1) seems to work but when doing setgamef(1,0) it crashes after saying "Welcome to my home. I take you on a guided " *crash*

vvs-:
I can't reproduce the crash itself. But all other speech is muted after that introduction.

Sur3:
True I tested on an other machine there it had the same problems as vvs- described, so probably something is wrong with the audio playback and my first test machine isn't tolerant about that.. xD

TeslaRus:
In my case I have engine stuck on start (but it is other bug)... debug does not help.
I use Al Soft (TeslsRus/OpenTomb build in Al library) - no audio problems

vvs-:
Not only that. Now I've got lock-up in introduction video. The stack backtrace shows that OpenAL code is the culprit. Most music triggered inside levels is not playing too indeed.

vvs-:
After last commit which removed threading from codec all issues still persist.

RICCIARDI-Adrien:
It seems to me that the Lara speech (like "Welcome to my home. I take you on a guided tour.") is played from an OGG file (029.ogg), is that right ?
However, this same speech is present in the level file (it's the sound ID 174 in my french version). Moreover, the level-provided speech is localized. Why don't play the level-provided speech instead of the OGG one ?

vvs-:
All speech in Lara Home level is played by their corresponding triggers and they use CD audio tracks.

BTW, trigger debug `r_triggers` doesn't show level triggers anymore after recent changes. Probably it using OpenGL API in some way which isn't compatible with every driver. May be black on black, or it assumed that background is pre-filled with some particular color, I don't know.

RICCIARDI-Adrien:
Some bugs described here should be fixed by pull request #595 merge. Could someone confirm that fixes are working for English and non-English games versions ? They work for French version.

vvs-:
I confirm - sound and video now work in English version.

Sur3:
Confirmed, Audio works and crash is also fixed.

PR #595, opened 2019-06-08
FMV fixes.

RICCIARDI-Adrien:
Following videos were crashing the game when trying to play them (it was due to unsupported audio format) :
* data/tr1/fmv/CAFE.RPL
* data/tr1/fmv/CANYON.RPL
* data/tr1/fmv/LIFT.RPL
* data/tr1/fmv/PRISON.RPL

Also, FMVs did not started from the beginning.

I'm trying to solve the bug that hangs the game when playing more than one FMV in the row, but you can merge this pull request if it is OK for you.

### DIFF
--- a/src/audio/audio_stream.cpp
+++ b/src/audio/audio_stream.cpp
@@ -49,7 +49,6 @@ void StreamTrack_Init(stream_track_p s)
 
 void StreamTrack_Clear(stream_track_p s)
 {
-    StreamTrack_Stop(s);
     s->buffer_offset = 0;
     s->linked_buffers = 0;
     if(alIsSource(s->internal->source))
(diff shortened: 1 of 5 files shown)


vvs-:
Right. They did actually work in the past but become broken at some time before #538.

RICCIARDI-Adrien:
Pull request is ready for merging.

TeslaRus:
Thanks for PR. I need some time to check it (I have little time).
Can not reproduce video play crash.

RICCIARDI-Adrien:
You're welcome.
I just checked and all videos that were crashing contain localized dialogs. All other videos do not have voice. Maybe my french version is 8-bit encoded but the english-one is 16-bit, like the voiceless videos ?

vvs-:
I have no crash either in my English version. But I have a lock-up playing any video for the second time. This is the same bug as in #538 where original reporter had non-reproducible crash on one of his computers.

Issue #600, opened 2020-05-16
Is the developer of this still here?


Nathidraws:
No activity after 2018..

RICCIARDI-Adrien:
It would be great to continue this awesome project !

vvs-:
That's not the real problem, anyone can fork it. But to be actually useful, any project should have a maintainer who will set real goals and responsibly dedicate efforts to achieve all these goals in foreseeable future. Among others such goals apparently should include actual documentation and regular builds for all supported platforms. There is no sense in yet another dead project.

RICCIARDI-Adrien:
Sure, there are already plenty of forks. What's cool with OpenTomb is that it is the original project !
Maybe all people wishing to participate could think of a roadmap and do some tasks until TeslaRus returns ?
For instance, @vvs- suggestion about regular builds may be achieved with GitHub Actions.

vvs-:
If indeed there were any people who wanted to work on this project they'd do something already. But unfortunately there is virtually no documentation, no standard Windows (or Mac OS) build environment and plenty of fundamental incompatibilities and bugs, e.g #91 or #417. There already were few attempts to set up nightly builds with Travis but nothing sustainable so far.

Of course, anyone is welcome to resolve all this.

Gh0stBlade:
The last commit was made September 2019. Have some patience.

RICCIARDI-Adrien:
As TeslaRus was building OpenTomb with Cygwin on Windows, it may be possible to create an automated Windows build using a GitHub Actions Ubuntu image with mingw. I already did that for another project, so I can give a try if the community is OK.

Another suggestion of something I should be able to do : rename repo `autoexec.lua` and `config.lua` to `autoexec_dev.lua` and `config_dev.lua`, so they are not loaded by default, and provide clean scripts for users so they do not get lost when starting the engine. This would imply setting the same keymap as original PC tomb raiders by default, as people may be more familiar with this mapping.
What do you think about these proposals ?

vvs-:
The development process of this project is not community driven - it follows benevolent dictator pattern. There was a short period of time when there was attempt to make decisions collectively, but that attempt failed miserably and the project lost all of its other developers. After return of TeslaRus only he led the project and made design decisions. Nobody else, except him and Lwmte could change global repository settings.

[...]

PR #601, opened 2020-05-18
Added Linux and Windows CI builds using GitHub Actions

RICCIARDI-Adrien:
Windows build should also be easier even for developing purposes.


TeslaRus:
try this AL internal, please https://github.com/TeslaRus/OpenTomb/tree/master/extern/al

RICCIARDI-Adrien:
OK.
What C files did you compile ? All the ones contained in `al` subdirectories ?
May I add a CMakeFiles.txt to `al` directory ?

TeslaRus:
yes, all files
May I add a CMakeFiles.txt to al directory ? - ok )

RICCIARDI-Adrien:
@TeslaRus, are you using MinGW on Cygwin ?
It seems that Windows threads have been removed from your OpenAL version, only pthreads are left.

TeslaRus:
yes, mingw has pthreads library (cut version, but enough for AL)

RICCIARDI-Adrien:
New OpenAL is now fully integrated. It is statically linked to avoid deploying one more DLL.

vvs-:
I've just checked the result: OpenTomb repository size has doubled because of the Windows binaries. I think that was not a good idea, because git is really bad at handling binaries. I'd split those away or this blow up could become a problem with time.

vvs-:
Reverted as per discussion in #600.

